PR #51, opened 2014-12-04
added `caldera_forms_mailer_nl2br` filter to prevent undesired `<br>` tags

diegoliv:
I had some issues with the mailer feature using the visual editor to build the body of emails with tables, where the `nl2br()` function was adding a `<br>` tag for each table line. This pull request keeps the original behavior, but gives the user a way to remove this behavior if is outputting undesired tags.


Desertsnowman:
This is a good option but I feel it relies to much on the filter being used. I think a better way would be to use the ```wpautop()``` that way the editor at least gets the proper formatting.
I could then add a pre and post filter to it.

PR #55, opened 2015-01-04
Replaced get_site_url with get_home_url for better sub-directory install compatibility

wpappdev:
get_site_url was pulling the sub-directory install into the preview form url resulting in a redirect loop.


Desertsnowman:
Thanks!

PR #77, opened 2015-02-22
Fix duplicate "$magic_tag =" in entry_token and add a {ua} tag to record HTTP_USER_AGENT 

sglfree:
Fix duplicate "$magic_tag =" in entry_token and add a new {ua} tag to record HTTP_USER_AGENT


Desertsnowman:
Thanks, well spotted.

PR #100, opened 2015-04-07
add readonly option to single line text

srt4rulez:
Here's another small edit I made for my own personal use. 

Just added an option for a "readonly" on the text input.


Desertsnowman:
Thanks. this is nice but not something that is vital for a capture form. I think it would be best to make this as an add-on rather. That way when you update, you don't need to fix the field.

PR #124, opened 2015-04-30
Fix undefined index: file

### LINKED ISSUE #123: Undefined index: file
When a file upload field is added to the form and `Attach a CSV version of the submission` is enabled, an undefined index is shown

PHP Notice:  Undefined index: file in /wordpress/wp-content/plugins/Caldera-Forms/classes/core.php on line 733

sumobi:
Fixes #123


Desertsnowman:
this shouldn't try attach an empty string but rather skip over if there's an error or the file is not there or doesn't exist. I made the change in branch 1.2.x. Thanks for pointing it out!

Issue #202, opened 2015-08-19
Ability To Set Min/ Max day in date picker


Shelob9:
Requested here: https://wordpress.org/support/topic/conditional-logic-with-dates?replies=1

soumyabg:
Just a temporary hack till this fature is included ,
* Go to caldera-forms/fields/date_picker/datepicker.php, 
* Search for 
```html
<input <?php echo $field_placeholder; ?> type="text" data-provide="cfdatepicker"
```
* Add this attribute inside the the <input>
```html
data-date-start-date="+1d" data-date-end-date="+2y"
```
* Save
Few examples of realtive dates:

Value | Relative date
------------ | -------------
0 | today
+1d | tomorrow
-1d | yesterday
-1m | last-month, this day
+1m | next-month, this day
-1y | last-yer, this day
+1y | next-year, this day

soumyabg:
I have forked the code and implemented the "Ability To Set Min/ Max day in date picker" 

https://github.com/soumyabg/Caldera-Forms

Requesting the community to check the code

Shelob9:
@soumyabg This looks very promising. It looks good to me, but I have not tested yet.

Can you please submit a pull request against the 1.2.x branch?

Desertsnowman:
+1 from me.

soumyabg:
I'll test it today, and confirm you guys later

soumyabg:
I have tested it, changed a little, now it is working pefectly.

Here is my pull request.
https://github.com/Desertsnowman/Caldera-Forms/pull/204

Screenshot
![5845449704472576](https://cloud.githubusercontent.com/assets/509036/9401765/ea3cf27a-47f1-11e5-8100-dc506fce17e2.png)

Shelob9:
Close via #204 

Thanks @soumyabg !

PR #204, opened 2015-08-21
Ability To Set Min/ Max day in date picker (The #202 Enhancement)

soumyabg:
Now, user can set a Start and End date range in the date picker.

The date range should in relative value like:

Value | Relative date
------------ | -------------
0 | today
+1d | tomorrow
-1d | yesterday
-1m | last-month, this day
+1m | next-month, this day
-1y | last-yer, this day
+1y | next-year, this day

If left empty, it means Infinity.

Screenshot:


Desertsnowman:
Thanks! I will test and confirm a few things. Much appreciated.

PR #207, opened 2015-08-22
Corrected an error in Date-picker module

soumyabg:
I actually made a stupid mistake by removing id and value fields which store and show given values to the fields, so I corrected it. Sorry.


soumyabg:
@Desertsnowman

PR #323, opened 2015-11-04
Add filters to autopopulate field options

JiveDig:
This allows use of the UI for the main settings, but use of the filter to further modify the get_posts or get_terms queries to allow more refined control of the field options.

### DIFF
--- a/classes/core.php
+++ b/classes/core.php
@@ -6,7 +6,7 @@
  * @package   Caldera_Forms
  * @author    David <david@digilab.co.za>
  * @license   GPL-2.0+
- * @link      
+ * @link
  * @copyright 2014 David Cramer
  */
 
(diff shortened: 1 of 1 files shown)


Desertsnowman:
wow, do you think you can re do this without the reformatting of spaces and new lines?
this commit indicates 193 additions and 190 deletions and I would prefer to keep it down to the exact changes

JiveDig:
Yeah i was just looking at this... not sure how that even happened. I thought I created a new branch directly from current-dev. I'll try again.

Desertsnowman:
its can happen if your code editor does auto formatting. I think.

JiveDig:
Just resubmitted, but I see the issue. In Sublime Text I have `"trim_trailing_white_space_on_save": true,`
So on save it automatically removes extra spaces after any characters on a line. It's def something for you to consider as well. I'll turn that off and resubmit again now.

PR #325, opened 2015-11-04
Resubmit PR 323 without trailing whitespace trim

JiveDig:
https://github.com/Desertsnowman/Caldera-Forms/pull/323

### DIFF
--- a/classes/core.php
+++ b/classes/core.php
@@ -1580,6 +1580,7 @@ class Caldera_Forms {
 						'order' => $field[ 'config' ][ 'order' ],
 						'orderby' => $field[ 'config' ][ 'orderby_post' ]
 					);
+					$args  = apply_filters( 'caldera_forms_autopopulate_post_type_args', $args );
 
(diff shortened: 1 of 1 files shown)


Shelob9:
@JiveDig I'm a little torn on this filter. Seems too downstream for me. The method its in is already hooked to `caldera_forms_render_get_field` so if you want a totally custom query to autopopulate the field options, why not use that?

@shelo9 he could, but then he as to copypaste all the formatting this method does form him...

@shelob9 Good point, that needs to be abstracted out a bit.

JiveDig:
@Shelob9 I agree, partially. It's slightly redundant since you can override the entire function for what's displayed for options. But in the test case for this, all I wanted to do was limit a CPT's select options to posts from the logged in author. My added filter would let me use the UI for all settings, then just filter to add `author` => get_current_user_id()` and i'm done.

Desertsnowman:
No, I like this. it means that auto-populating can be natively filtered father than completely overridden.

PR #328, opened 2015-11-05
removed double slashes for assets that are actually valid but not bes…

rene-s:
…t practice and cause problems in some cases.

Background story: I have a client that employs a certain single-sign-on system that bails out on URLs with double slashes. In order to make this plugin work for that client, I have removed the double slashes.

Regards
rs


Desertsnowman:
Actually that was a bug that I was supposed to fix. Thanks for spotting it!

PR #332, opened 2015-11-08
Use default WordPress collation for database tables.

ChrisWiegman:
Hard-coding UTF8, particularly after WordPress Core's move to UTF8mb4 can cause issues in some database setups. This will allow the tables to use the default DB collation with a fallback to the default collation for the database itself without risk of introducing anything that could cause issues.


Shelob9:
Would fix #333 

Thanks @ChrisWiegman

Desertsnowman:
Oh look, a facepalm moment! -1 to me.
much appreciated!

PR #450, opened 2016-03-05
export of entries to CSV  - missing values with checkboxes bug

owen546:
CSV export of entries has blank fields for checkboxes when only one is ticked.

In the file classes/admin.php

Line 1434
$row[$key] = $row[$key][0];

Will not work for some checkboxes (when only one value is ticked) because in that case $row[$key][0] is an associative array not a numerical one.

So, to work in all cases, I replaced Line 1434 with the following

reset($row[$key]);
$sub_arr_key = key($row[$key]);
$row[$key] = $row[$key][$sub_arr_key];


owen546:
fixed in 1.3.4.1

Issue #502, opened 2016-03-30
Feature Request: Form Processor Add Error Message attached to Field Option


brandbrilliance:
Currently a form processor only targets the field itself with the error shown at the bottom of all the options (fixed in #430). However, it would be nice to allow a target even deeper to the actual option label itself. In this case, you could check the array and if it contains an array, and the option matches the key set in the array, it could highlight **only ** the error option (instead of all the options) and add the error message to just that option (usually below).

**Suggested New Code using option array as well:**
```
Array ( 
[type] => error 
[note] => Seat Limit reached for Location: Cape Town (0 free seats)
[fields] => Array ( 
     [fld_3597343] => array (
          [opt1697861] => No Free Seats for Location: Cape Town (0 free seats) 
))
```

![options-error-suggested](https://cloud.githubusercontent.com/assets/1362479/14141006/b29f5c42-f67c-11e5-8f97-fb556b84bb2d.jpg)

PR #509, opened 2016-03-30
Update ajax-core.js to support field option errors 

brandbrilliance:
I've added my hack code, probably not 100% correct for all field types which will check the processor data error string and if it's got a plain string, it's only a field error, but if it contains an object, it will iterate the object key/value pairs and find the matching option value and add the error message below the relevant option and flag only that option as has-error. the double looping means you'll be able to set multiple field option errors at the same time (if present in the data error return string).  (Feature Request #502)

### DIFF
--- a/assets/js/ajax-core.js
+++ b/assets/js/ajax-core.js
@@ -82,21 +82,77 @@ jQuery(function($){
 				if(obj.data.fields){
 
 					for(var i in obj.data.fields){
-						var field = obj.params.trigger.find('[data-field="' + i + '"]'),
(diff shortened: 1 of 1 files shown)


Shelob9:
@brandbrilliance Thanks for your PR. I'm sure @Desertsnowman will want to review this.
Can you please re-submit this against the current-dev branch? We don't merge anything against the current-stable branch.

brandbrilliance:
Not Like I know how to do that, not a professional coder. LOL.

PR #514, opened 2016-03-31
Update to support field option errors

brandbrilliance:
I've added my hack code, probably not 100% correct for all field types which will check the processor data error string and if it's got a plain string, it's only a field error, but if it contains an object, it will iterate the object key/value pairs and find the matching option value and add the error message below the relevant option and flag only that option as has-error. the double looping means you'll be able to set multiple field option errors at the same time (if present in the data error return string). (Feature Request #502)

### DIFF
--- a/assets/js/ajax-core.js
+++ b/assets/js/ajax-core.js
@@ -82,21 +82,77 @@ jQuery(function($){
 				if(obj.data.fields){
 
 					for(var i in obj.data.fields){
-						var field = obj.params.trigger.find('[data-field="' + i + '"]'),
-							wrap = field.parent();
-						if( wrap.is('label') ){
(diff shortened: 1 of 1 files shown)


Desertsnowman:
@brandbrilliance this is great, but it's really the wrong way to do it.
The way I'm currently doing it is also wrong as this kind of thing needs to be handled by the field itself.

reason being is that since the field structure is filterable and can change depending on what the implementation requires. So we can't rely on a default structure.

I do have a new way I'll get setup though but for now, I'll keep it the way it is now.
You can use your own method, btw. 
If you put that function in the theme or somewhere then edit the form and go to general settings -> Custom Callback. then place the function name there.

PR #558, opened 2016-04-22
Add date support in calculation fields

pboissonneault:
Add date support in the calculation fields.

Forget the ui/edit.php


Shelob9:
I am closing this as it can't be merged any more. Mega point deductions to me for never merging/ reviewing this, I feel bad, but it came in at a bad time and then got stale. We need to have it against 1.5.x, but if it could fix #908 that would have been awesome.

fernigs:
how do i add date support? where? 
Thx

PR #833, opened 2016-09-26
default value for orderby_post for autopopulate post_type

kibarda:


Shelob9:
@kibarda  Thanks for this PR and for catching the mistake (I made issue #836 about it). But we do not accept pull requests against master branch. We only change master branch when the release is ready.

Can you please re-submit against the 1.4.x branch ?

PR #1000, opened 2016-12-09
New font icon created and added to the code

isotrope:
The new icon font


classes/admin.php
```diff
@@ -1011,7 +1011,7 @@ public function register_admin_page(){
 		// get current user
 		if( current_user_can( Caldera_Forms::get_manage_cap() ) ){
 
-			$this->screen_prefix[] = add_menu_page( __('Caldera Forms', 'caldera-forms' ), __('Caldera Forms', 'caldera-forms' ), Caldera_Forms::get_manage_cap(), $this->plugin_slug, array( $this, 'render_admin' ), 'dashicons-cf-logo', 52.81321 );
+			$this->screen_prefix[] = add_menu_page( __('Caldera Forms', 'caldera-forms' ), __('Caldera Forms', 'caldera-forms' ), Caldera_Forms::get_manage_cap(), $this->plugin_slug, array( $this, 'render_admin' ), 'dashicons-caldera-logo-2016', 52.81321 );
```
Shelob9: @isotrope I don't like adding `-2016` here. We should be erasing all traces of old icon. There are some other places that dashicon is used, we want same CSS to magically become on brand. You are a branding magician today.

isotrope: I will try very soon.

Issue #1027, opened 2016-12-16
After upgrading to 1.4.6.1, summary magic tag producing one-letter values


cbomkamp:
I installed Caldera Forms on my site on October 30, so whatever was the latest version at that time is what I had previously. I setup one or two forms and some auto-responders with just the {summary} tag in the body. All was working fine at that point. I'm using the Mailgun plugin to send mail through WP.

I upgraded to 1.4.6.1 yesterday and now the emails come through looking like this:

> **First Name**
> C
> 
> **Last Name**
> B
> 
> **Email Address**
> c
> 
> **Your Age (must be 21 or older)**
> 2
> 

I'm only getting the first character of each field value. It doesn't matter what I enter in the fields. I always just get 1 character.

The data shows correctly in the entries list in the admin panel.

Also, I can use the individual field tags in the processor and they will show up fine. For example, if I put in %first_name%, I get the full first name value.

cbomkamp:
Just noticed this was posted on the support forums:

https://wordpress.org/support/topic/summary-magic-tag-problem/

Shelob9:
@cbomkamp Thanks for using Caldera Forms. I'm not seeing this in any of my test environments. If possible, can you please send me an export file for your form (Josh@CalderaWP.com) also can you try it with no other plugins active and with the default theme?

crataz:
Same problem for me, I sent you an email.

cbomkamp:
@Shelob9 I sent you an email.

Shelob9:
@crataz @cbomkamp Thanks for sending those exports, I will test later today. It would be very helpful to know what PHP versions both of you are using. Please look in the Caldera Forms support submenu, under the debug info tab.

cbomkamp:
WordPress Version: 4.6.1

PHP Version: 5.2.17

crataz:
WordPress Version: 4.7

PHP Version: 5.3.29

MySQL Version: 5.5.52

Server Software: Apache

Shelob9:
Thank you both for that info. I'll see if your out of date PHP could be the
cause.

I strongly recommend both of you update to PHP 5.6. Your versions of PHP
are out of date, slow and potentially insecure.

cbomkamp:
Any progress?

cbomkamp:
[...]

PR #1086, opened 2017-01-04
Fixing how PHP 5.3 checks array keys

vinnycaldera:
fix CalderaWP/Caldera-Forms#1027


Shelob9:
@vinnycaldera While this does fix the issue, testing in PHP 5.2 - it is also causing a warning `array_key_exists() expects parameter 2 to be array, string given in /app/public/wp-content/plugins/Caldera-Forms/classes/magic/summary.php on line 99`

Issue #1087, opened 2017-01-04
&nbsp; should not be used in radio field


Troglobitten:
This plugin uses &nbsp; in it's output on radio type input fields. These have no use except for spacing the separate radio fields. Spacing should be left to the stylesheet.

Shelob9:
I agree. Easy fix here: https://github.com/CalderaWP/Caldera-Forms/blob/master/fields/radio/field.php#L37

Shelob9:
Close via #1093

PR #1092, opened 2017-01-06
Removed &nbsp; from radio and checkbox type inputs

Troglobitten:
These have no use except for spacing the seperate radio and checkbox
fields. Spacing should be left to the stylesheet.

See #1087


Shelob9:
@Troglobitten Thank you for the PR.

We do not accept pull requests against master.l Can you please, resubmit this against the 1.4.x branch?

PR #1093, opened 2017-01-06
Removed &nbsp; from radio and checkbox type inputs

### LINKED ISSUE #1087: &nbsp; should not be used in radio field

Troglobitten:
These have no use except for spacing the seperate radio and checkbox
fields. Spacing should be left to the stylesheet.

fixed #1087 

Hopefully this is on the correct branches this time, I'm new to this whole committing thing so thank you for your patience

### DIFF
--- a/fields/checkbox/field.php
+++ b/fields/checkbox/field.php
@@ -28,7 +28,7 @@
 				<?php if(empty($field['config']['inline'])){ ?>
 				<div class="checkbox">
 				<?php } ?>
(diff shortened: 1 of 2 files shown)


Shelob9:
@Troglobitten  Thanks for this. I am hoping to do a 1.4.8 release next week or soon after. This will include that.

Issue #1061, opened 2016-12-21
Scroll To Top On Submit - Form settings repeated 2 times


paaljoachim:
Hey

Scroll To Top On Submit in the Form settings is repeated twice.

![screen shot 2016-12-21 at 11 20 26](https://cloud.githubusercontent.com/assets/5323259/21385650/a300763a-c76f-11e6-9733-97311f0a0c3a.png)

Shelob9:
Second one should be deleted

Shelob9:
Close via #1104

PR #1104, opened 2017-01-11
Removed repetition of "Scroll To Top On Submit" Section

faisal-calderalabs:
I have removed repetition of **Scroll To Top On Submit** section from email setting in admin panel.


Shelob9:
#1061

PR #1115, opened 2017-01-17
Patch 1

101010f:


Shelob9:
@101010f You need to submit this against 1.5.x branch, not master, that's why you're getting so much extra stuff.

PR #1117, opened 2017-01-17
Removed 1px min-height from rows

101010f:
I proposed a radio button rearrangement in here as well if you're interested. I use it to style checkboxes and radio buttons. 
Closed the last pull request. I'm just getting the hand of this thing. Cheers.

### DIFF
--- a/assets/css/caldera-grid.css
+++ b/assets/css/caldera-grid.css
@@ -634,7 +634,6 @@
 .caldera-grid .col-md-12,
 .caldera-grid .col-lg-12 {
   position: relative;
-  min-height: 1px;
   padding-left: 7.5px;
   padding-right: 7.5px;
 }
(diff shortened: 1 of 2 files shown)


Shelob9:
@101010f I hate to be difficult beacuse I super-appreciate the contribution but pull requests have to be for one issue. I'm not just saying that beacuse it's good work flow, but beacuse you have two legit, but discreet issues.

The correct solution for the problem you have with radio is to use a `<fieldset>` tag, like we have done in admin UI.

If this was just 194bbc4 I would merge.

101010f:
OK I would resubmit with just 194bbc4. I had the input thing from a previous update and didn't know how to delete it. Sorry about that. I'll try and figure this out and submit a single issue.

Shelob9:
@101010f You have to keep each PR to a single branch. If you make a commit to your 1.5.x branch and then submit PR, it looks good, but then if you make another commit to that branch before I merge it, the PR will be for both commits.

PR #1119, opened 2017-01-17
Remove 1px min row height

101010f:
Hope I nailed it this time. xD


Shelob9:
@101010f This is good, I will merge later today. Thank you for your help and for resubmitting and for resubmitting ;) I hope you understand why I needed it this way and will submit the other PR separately (see #1126).

I just need to make sure I do it when I have a moment to recompile CSS right away.

Issue #1186, opened 2017-01-31
HTML field not working


andrewlimaza:
Version 1.5-b-2

HTML form not rendering any HTML (i.e. `<h2>Some simple header</h2>`)

No errors in console. 

Not sure if this is site specific, PHP 5.5

Shelob9:
@andrewlimaza Does it never render anything? IE if you don't use HTML, does that work?

andrewlimaza:
@shelob9 yea even plain text not wrapped in any HTML tags doesn't show.

andrewlimaza:
@Shelob9 the div for the HTML is outputting to the browser, just no data for it.

andrewlimaza:
Update: issue seems to be with the sanitization.

andrewlimaza:
@Shelob9 PR -> https://github.com/CalderaWP/Caldera-Forms/pull/1191

Shelob9:
Close via  #1191

PR #1191, opened 2017-02-01
Remove 'html' from forbidden_tags array

andrewlimaza:
Remove 'html' from forbidden_tags array. Fixes content not displaying on frontend when using HTML field.


Shelob9:
Fixes #1186

geoffreygordon:
how do you implement this, I have the same issue

Issue #1327, opened 2017-02-28
Edit entry issue: Text Area shows blank once editing data


andrewlimaza:
Entries correctly capturing data - however when you start to edit the entries the textarea shows up as empty even though in the view entry it had data.

Tested on version 1.5.0.4

Vrhovc:
I have observed this issue also..
I populate form with caldera_forms_render_pre_get_entry filter, and all of the other fields get populated as they should, only textarea field does not, it stays empty. Checking the returned variable has value as it should.

Shelob9:
Close via #1413

PR #1413, opened 2017-03-16
Fix for issue #1327 (textarea value blank on edit)

francois-parodi:
Fix for issue #1327: Text Area shows blank once editing data (bug)
The value of a text area was always set to its default value, even when editing.


Shelob9:
Thanks @francois-parodi

jlangley:
@Shelob9 It looks like this will fix the textarea piece of #1520 but it didn't make it into release [1.5.0.7](https://github.com/CalderaWP/Caldera-Forms/milestone/36) because a CI check failed on 3/16.

https://travis-ci.org/CalderaWP/Caldera-Forms/builds/211875580

PR #1425, opened 2017-03-21
$entryid instead $entry_id detected in classes\core.php (BUG?) 

ESPONI:
near line 3061 in classes\core.php
$wpdb->insert( $wpdb->prefix . 'cf_form_entries', $new_entry );
$entryid = $wpdb->insert_id;

After that, $entry_id is used so NULL value is passed to all actions/filters.


Shelob9:
@ESPONI  Thanks for the pull request, but I can't merge this. The 1.5.x branch is no longer in use. Right now we are using master for bugs, like this, and develop for features.

Issue #1520, opened 2017-04-19
Paragraph Textarea and Star Rating Do Not Use Default Values When Editing Form Entries


jlangley:
## What Version Of Caldera Forms, WordPress and PHP Are You Using?
WordPress Version: 4.7.3PHP Version: 5.6.30MySQL Version: 5.6.28Caldera Forms Version: 1.5.0.8WP_DEBUG:

## Does Your Issue Presist When You Disable All Other Plugins and Switch To The Default Theme?
I'm on a production build, so not able to easily test this. If you cannot replicate my issue, I'll stand up a sandbox and try.

## What Is The Unexpected Beahviour?
When attempting to edit a submitted form that has "paragraph textarea" and "star rating" controls, no default value is set on those controls. This makes editing the previously submitted forms impossible because the new (blank/standard) defaults will override the previously submitted values.

## What PHP Errors Have You Logged While Reproducing This Bug?
No Caldera Forms errors. When I turn on PHP Debug logging, I see this:

> Strict Standards: Non-static method OLT_Tab_Shortcode::add_shortcode() should not be called statically in /home/gardenth/public_html/wp-content/plugins/tabs-shortcode/tabs-shortcode.php on line 79
> 
> Strict Standards: Non-static method OLT_Tab_Shortcode::has_shortcode() should not be called statically in /home/gardenth/public_html/wp-content/plugins/tabs-shortcode/tabs-shortcode.php on line 67
> 
> Strict Standards: Non-static method OLT_Tab_Shortcode::add_shortcode() should not be called statically in /home/gardenth/public_html/wp-content/plugins/tabs-shortcode/tabs-shortcode.php on line 80
> 
> Strict Standards: Non-static method OLT_Tab_Shortcode::has_shortcode() should not be called statically in /home/gardenth/public_html/wp-content/plugins/tabs-shortcode/tabs-shortcode.php on line 67

## What JavaScript Errors Have You Seen While Reproducing This Bug?
No javascript errors.

Shelob9:
The paragraph field is now fixed via #1538

Shelob9:
Close via  3b86fc1 (paragraph issue) and 		ececb0b (default for stars)

PR #1521, opened 2017-04-20
Fixed star rating edition (issue #1520)

francois-parodi:
When editing a form with a star rating, the correct number of stars is
shown instead of the default number of stars.


Shelob9:
@francois-parodi This looks great, thanks for the pull request. If you sbmit this against master, I'll merge it and include it in both 1.5.0.9 and 1.5.1, if I merge this against develop, no one will get it until 1.5.1 is released.

Issue #1610, opened 2017-05-24
Duplicate column 'status' error on plugin activation


shawnhooper:
## Is This A Feature Request?

No

## What Version Of Caldera Forms, WordPress and PHP Are You Using?

Caldera Forms 1.5.0.9
PHP 7.0.15-0ubuntu0.16.04.4 (cli)

## Does Your Issue Presist When You Disable All Other Plugins and Switch To The Default Theme?

Did not test

## What Is The Unexpected Beahviour?

I installed the Caldera Forms plugin using WP-CLI:

`wp plugin install caldera-forms --activate`

and the error below appeared. 

## What PHP Errors Have You Logged While Reproducing This Bug?

WordPress database error Duplicate column name 'status' for query ALTER TABLE `wp_cf_form_entries` ADD `status` varchar(20) NOT NULL DEFAULT 'active' AFTER `datestamp`; made by include('/var/www/html/consultants/vendor/wp-cli/wp-cli/php/wp-cli.php'), WP_CLI\Runner->start, WP_CLI\Runner->_run_command, WP_CLI\Runner->run_command, WP_CLI\Dispatcher\Subcommand->invoke, call_user_func, WP_CLI\Dispatcher\CommandFactory::WP_CLI\Dispatcher\{closure}, call_user_func, Plugin_Command->install, WP_CLI\CommandWithUpgrade->install, Plugin_Command->activate, activate_plugin, do_action('activate_caldera-forms/caldera-core.php'), WP_Hook->do_action, WP_Hook->apply_filters, Caldera_Forms::activate_caldera_forms

WordPress database error Duplicate key name 'status' for query CREATE INDEX `status` ON `wp_cf_form_entries` (`status`);  made by include('/var/www/html/consultants/vendor/wp-cli/wp-cli/php/wp-cli.php'), WP_CLI\Runner->start, WP_CLI\Runner->_run_command, WP_CLI\Runner->run_command, WP_CLI\Dispatcher\Subcommand->invoke, call_user_func, WP_CLI\Dispatcher\CommandFactory::WP_CLI\Dispatcher\{closure}, call_user_func, Plugin_Command->install, WP_CLI\CommandWithUpgrade->install, Plugin_Command->activate, activate_plugin, do_action('activate_caldera-forms/caldera-core.php'), WP_Hook->do_action, WP_Hook->apply_filters, Caldera_Forms::activate_caldera_forms

[...]

PR #1611, opened 2017-05-24
#1610 Fix status column creation error on activation

shawnhooper:
Fixes issue #1610.  The $fields array was always empty, so activation of the plugin on a new install would always attempt to create the field and the key a 2nd time.


Shelob9:
Thanks @shawnhooper

PR #1616, opened 2017-05-24
Add interface & base class to support future CRM processors

shawnhooper:


Shelob9:
@shawnhooper I don't think the methods in this class are needed, since they exist in the class they are extending.

shawnhooper:
@Shelob9 Yup, good point.  That was a copy/paste from the newsletter processor.  No actual CRM processor to test this against yet, just wanted the class name there to extend from.

Shelob9:
#1619

Shelob9:
@shawnhooper The @since should be 1.5.1 otherwise Taco Emoji

shawnhooper:
@Shelob9 You may proceed with Taco emoji.

Issue #1638, opened 2017-06-04
Add user:roles as a magic tag


agileware:
It would be incredibly useful to have user:roles available as a magic tag in Caldera Forms. This can then be used:
1. in conditions to perform certain actions based on the user having one or more roles
2. store the user roles of the submitting user

Shelob9:
@agileware The users add-on provides form and field visibility based on logged in state and role: https://calderaforms.com/downloads/caldera-forms-users-add/ It doesn't do a magic tag, but that should be added there: https://bitbucket.org/calderadev/cf-users/issues/3/add-user-role-magic-tag

agileware:
@Shelob9 That's a bummer dude :( Would you reconsider this decision?

Shelob9:
@agileware I can't take paid functionality and add it to the free plugin, that doesn't make sense. The Users add-on sells very well for us, if we don't sell add-ons then we don't have a business and then I can't keep making Caldera Forms and Caldera Forms related things. I like making Caldera Forms and Caldera Forms related things.

PR #1639, opened 2017-06-04
Add user:roles as a magic tag

agileware:
Add user:roles as a magic tag, PR for #1638


agileware:
Apologies for the PR against master branch, if that's wrong. Did try against the develop branch initially, but the PR was huge.

PR #1645, opened 2017-06-07
Fix for magic tags

danisiz:
This fix processor's magic tags from leave working solving a recent code change mistake.


Shelob9:
Thanks @danisiz !

Issue #1659, opened 2017-06-11
No way to set default dropdown option using magic tags


bensteers:
I am using Caldera as essentially a profile editor. I need to set the form values using data pulled from my database via magic tags. For text fields, that is easy because I can just use a magic tag in the default value argument, but for dropdowns (and other selects) the default value can not be set using magic tags. I realize that the placeholder argument can take magic tags, but the value is set to be blank so that doesn't really help me.

One solution would be to add a text box to the Add Option section of the edit panel which could be matched against the values of the options. 

Another solution could be to just check the placeholder value against all of the option values and if it matches any of the option values, select that, otherwise use the existing functionality.

I can submit a pull request, but I was wondering which method would be preferred, or if I am missing anything. If neither, I can just patch it with hooks, but it would be more robust if it were part of the default functionality.

bensteers:
I'm thinking the first option could be ideal, especially if it is added to all 'selects'. It would be great if I could specify a default US state from the state dropdown.

Shelob9:
@bensteers Adding a default option, that was matched tag enabled makes a lot of sense. If you put in a pull request against master, I will merge it.

Shelob9:
Close via #1660 etc.

PR #1660, opened 2017-06-12
Added default value text boxes for select options

bensteers:
The fields affected are:
- dropdown
- checkbox
- radio
- select2
- toggle switch

Also, I renamed the "default" field slug in the Add Options section to "default_option" because it is more consistent with the usage of the default field elsewhere and it made implementing the default text field much easier. 

Just a note: checkboxes and select2 fields can't take multiple default values. Perhaps the easiest thing to do would use new line delineated textareas. Before that gets done, multiselect fields still use radio buttons to select the default value meaning that only one default can be chosen. These should be changed to checkboxes.

### DIFF
--- a/fields/checkbox/config_template.php
+++ b/fields/checkbox/config_template.php
@@ -6,3 +6,11 @@
         </label>
 	</div>
 </div>
+<div class="caldera-config-group">
+	<label for="{{_id}}_default">
+		<?php esc_html_e('Default Value'); ?>
+	</label>
+	<div class="caldera-config-field">
+		<input type="text" id="{{_id}}_default" class="block-input field-config magic-tag-enabled" name="{{_name}}[default]" value="{{default}}" placeholder='Matched against values'>
+	</div>
+</div>
--- a/fields/checkbox/field.php
+++ b/fields/checkbox/field.php
@@ -10,6 +10,18 @@
 			$parsley_req = 'data-parsley-required="true" data-parsley-group="' . esc_attr( $field_id ) . '" data-parsley-multiple="' . esc_attr( $field_id ). '"';
 		}
 
+		/////// Note this needs to be updated in order to handle array field_values !!
+		// If the field value set doesn't exist, set it back to null
+		if( !empty($field['config']['option']) ){
+			$option_values = array_map( function($v){ 
+				return isset($v['value']) ? $v['value'] : $v['label']; 
+			}, $field[ 'config' ][ 'option' ] );
+
+			if(!in_array($field_value, $option_values)){
+				$field_value = null;
+			}
+		}
+
 		if(!empty($field['config']['option'])){
 			
 			// If default exists and val doesn't, set it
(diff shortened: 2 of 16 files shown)


Shelob9:
#1659

Shelob9:
Thanks @bensteers this is really good. I'm going to merge and then make a few minor changes, want this in 1.5.1 (should be out this week.)

Shelob9:
@bensteers Three things I'm changing - mentioning FYI and also beacuse you might have a concern:

1) Fixing translation strings in UI. Have to have everything translatable in our text domain.
2) Redoing your use of array_map with annonymous function. Sadly, PHP 5.2 is a thing we support.
3) I'm keeping `default` not `default_option` beacuse wile `default_option` is semantically better, but I'd rather be consistent and always use `default` .

Shelob9:
So here is what went wrong. PR switched config.default to config.default_option, which isn't backwards compatible -- IE we'd loose saved settings. But my solution created two new problems:

1) We have two settings called config.default_option
2) Logic addresses default by Option value or label, but config.option is saved as option index (opt12345...)

*curses

bensteers:
Hey, sorry I'm getting back to you so late. So what if, on update, we just run a small migration script that moves over all  `field['default']` settings to `field['default_option']`? 

Addressing each of the problems:
1. What is the other setting that you're referring to? Did I accidentally overwrite another setting? I'm not sure if I understand.
2. Right but that shouldn't be a problem because we just extract the value and match it with the option values as they are being output, correct?

Here's a quick non-functional example of what we could do. I believe this could resolve the issue?
```php
$CF_VERSION_TO_MIGRATE = '1.5.1'; // or whichever version you update this
function caldera_forms_update_compatibility(){
	// Check plugin version
	$plugin_data = get_plugin_data('caldera_forms');
	if(strcmp($plugin_data['Version'], $CF_VERSION_TO_MIGRATE) >= 0 && get_option('caldera_forms_has_migrated_default_option')){

		// Get form objects
		$forms = array();
		foreach(Caldera_Forms_Forms::get_forms() as $form_id)
			$forms[$form_id] = Caldera_Forms_Forms::get_form($form_id);

		// Migrate default -> default_option
		$fields_affected = array('dropdown', 'checkbox', 'radio', 'select2', 'toggle');
		foreach($forms as $form_id=>$form){
			foreach($form['fields'] as $field_id => $field){
				if(in_array($field['type'], $fields_affected)){
					// Update field
					$field['config']['default_option'] = $field['config']['default'];
					$field['config']['default'] = null;
				}
			}
			// Save somehow
			//save_form($form_id, $form);
		}

		// Store a reference saying you completed this process
		update_option('caldera_forms_has_migrated_default_option', 1);
		// OR...
		// Store this in its own file and just delete it after it is done
		unlink(__FILE__);
	}
}
add_action('plugins_loaded', 'caldera_forms_update_compatibility');
```

Issue #1733, opened 2017-07-06
Modal Overflowing Incorrectly on Desktop View


YenDiesel:
## Do You Need Immediate Help?

NO

## Is This A Feature Request?

NO

## What Version Of Caldera Forms, WordPress and PHP Are You Using?

Short Debug Information

WordPress Version: 4.8PHP Version: 7.0.19MySQL Version: 5.5.5Caldera Forms Version: 1.5.2.1WP_DEBUG:

## Does Your Issue Presist When You Disable All Other Plugins and Switch To The Default Theme?

YES

## What Is The Unexpected Beahviour?
* Please explain what you did, what you expected to happen, and what happened instead.

I have added a contact form to my [contact page](https://east-coast-it.co.uk/contact-us/) 

When you test the form & hit the contact button, the modal window opens but is overflowing across the entire page when using a browser.

Mobile & Tablet are both unaffected

## What PHP Errors Have You Logged While Reproducing This Bug?
* See: https://calderaforms.com/2016/05/wordpress-debug-logging/

NONE

## What JavaScript Errors Have You Seen While Reproducing This Bug?
* See: http://webmasters.stackexchange.com/questions/8525/how-to-open-the-javascript-console-in-different-browsers

NONE

Shelob9:
@YenDiesel Thanks for using Caldera Forms and for pointing out this bug.

Looks like the solution is:
```
.caldera-modal-body 
.caldera-grid .row {
    max-width: 100%;
}

```

@New0 Can you test this one please? Do you see a better option?

Shelob9:
Close via  11434bc

PR #1744, opened 2017-07-10
.row max-width set to 100%

New0:
Avoid themes or plugins to load new .row width


Shelob9:
Fixes #1733

@New0 Please include issue number in commit messages and PR.

PR #1804, opened 2017-07-28
i18n syntax and typos

pedro-mendonca:


Shelob9:
Thanks @pedro-mendonca  !

Issue #1836, opened 2017-08-07
FEATURE: UI option for reply to in auto-responder


Shelob9:
BTW You can do this with the caldera_forms_autoresponse_mail filter

https://calderaforms.com/doc/caldera_forms_autoresponse_mail/

Shelob9:
#1816 and #1815  are related

Shelob9:
Close via  #1905

Issue #1814, opened 2017-08-01
FEATURE REQUEST: Customize Names of Processors


mathetos:
When using multiples of the same processor it would make it easier to manage them if the names could be customized.

![image](https://user-images.githubusercontent.com/930724/28841137-14c195d6-76ae-11e7-9660-bdaf4cffc20a.png)

Shelob9:
Close via  #1905

PR #1905, opened 2017-09-02
Refactor autoresponder

### LINKED ISSUE #1815: FEATURE REQUEST: Add BCC field to Auto Responder Processor
The Auto Responder is the best way to implement conditional email notifications (as far as I can tell). But in order for it to be as effective as the standard Email Notification, it should have at least the same features. Right now adding the BCC would be very helpful.

Shelob9:
Fixes #1815 #1814 #1836


Shelob9:
Also fixes #1906

Issue #807, opened 2016-09-06
Checkbox/radio html markup


101010f:
Is there a reason the checkbox or radio is nested in the label tag? This makes it imposible to style with css. Is it possible to have it like:

`<div>
<input type="checkbox"></input>
<label></label>
</div>`

and same for radioboxes ?

Shelob9:
This was annoying me the yesterday while working on our new site. I'd like to get it improved in 1.5. 

You can use the caldera_forms_render_field_file filter to make your own file for rendering a file type if you need to. https://calderawp.com/doc/caldera_forms_render_field_file/

But if you're going to do that, you should just submit a pull request (against the 1.5.x branch) for all to have.

101010f:
I'm not sure how to use the _caldera_forms_render_field_file_ since I'm not a programmer and I'd mess things up.
What did work though was pasting the input tag in checkbox/field.php before the input tag. Would this mess anything else in the plugin?

mrksdiehl:
Related to this one:
Is there a specific reason for the trailing **&nbsp;** in fields/checkbox/field.php on line 31 after the <label> tag? If the checkbox is without label, the whitespace will add a margin on the bottom that should not be there.

PR #1915, opened 2017-09-05
Refactor checkboxes and radios to use better markup

Shelob9:
Goal of this project -
```
<fieldset>
	<legend>Field Label</legend>
	<div>
		<input type="checkbox" value="option-value-1">
		<label>Option label 1</label>

		<input type="checkbox" value="option-value-2">
		<label>Option label 2</label>
	</div>
</fieldset>
```
#807


pascalknecht:
@Shelob9 Would love to have this feature. We always overwrite the field to have this markup

pascalknecht:
@Shelob9 What ist the status of this feature? Will it be included in the next release?

PR #1923, opened 2017-09-06
re-enable dropdown after bulk action - fixes #1833

### LINKED ISSUE #1833: Entries' Bulk Action dropdown remains disabled unless page is refreshed.
This occurs when managing entries of a form:
- At 'Active' tab, I select multiple entries, select a bulk action to move to trash and apply.
- Bulk Action dropdown becomes disabled
- I then go to the 'Trash' tab, I select multiple entiries, but Bulk Action dropdown remains disabled, and I can't apply anything.

Seems like once Bulk Action is applied once, it remains disabled no matter which form or tab I go to.

Using the following versions:
WordPress Version: 4.8.1
PHP Version: 5.6.31
Caldera Forms: 1.5.4

blunket:
after applying a bulk action, the bulk action dropdown would remain disabled until refresh. this patch re-enables it properly

### DIFF
--- a/ui/entries/scripts_templates.php
+++ b/ui/entries/scripts_templates.php
@@ -274,6 +274,8 @@ if( ! defined( 'ABSPATH' ) ){
 
 				row_items.animate({"opacity": .4}, 500);
 				$.post(ajaxurl, data, function(res){
+					action.val('').prop('disabled', false);
+					bulkCheck.prop('checked', false);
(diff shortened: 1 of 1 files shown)


blunket:
Hi @Shelob9 ,
I apologize if I did something wrong in contributing. I am newer to open source. Could you explain if so? Thanks.

Issue #1968, opened 2017-09-23
Compatibility with litespeed cache plugin (with solution inside)


hailite:
Please check this wiki for the compatibility improvement:

https://www.litespeedtech.com/support/wiki/doku.php/litespeed_wiki:cache:lscwp:esi_sample

LSCWP v1.2.4+ will be released this week. If the compatibility issue can be fixed, we can add your plugin into our compatible plugin list.

hailite:
I fixed it by https://github.com/CalderaWP/Caldera-Forms/pull/1971

Please check the pull request

Shelob9:
@hailite I can't find your email address. I was wondering if I could ping you about an issue? I am Josh@CalderaWP.com

hailite:
Yes I am here. About the nonce expiration issue, the temp solution is here:

If you want to set it on your side, you can change this part in render/nonce.php line 97:

```
		if ( ! $from_esi ) {
			if ( method_exists( 'LiteSpeed_Cache_API', 'esi_enabled' ) && LiteSpeed_Cache_API::esi_enabled() ) {
				if ( method_exists( 'LiteSpeed_Cache_API', 'v' ) && LiteSpeed_Cache_API::v( '1.2.4' ) ) {
					$params = array( 'form_id' => $form_id ) ;
					return LiteSpeed_Cache_API::esi_url( 'caldera_forms', 'Caldera Forms', $params ) ;
				}
			}
		}
```

to 

```
		if ( method_exists( 'LiteSpeed_Cache_API', 'esi_enabled' ) && LiteSpeed_Cache_API::esi_enabled() ) {
			if ( method_exists( 'LiteSpeed_Cache_API', 'v' ) && LiteSpeed_Cache_API::v( '1.2.4' ) ) {
				if ( ! $from_esi ) {
					$params = array( 'form_id' => $form_id ) ;
					return LiteSpeed_Cache_API::esi_url( 'caldera_forms', 'Caldera Forms', $params ) ;
				}
				else {
					LiteSpeed_Cache_API::set_cache_private() ;
					LiteSpeed_Cache_API::set_cache_no_vary() ;
				}
			}
		}
```

This will make the ESI block use private TTL setting which is 1800s by default. Should be able to fix this issue.

hailite:
@Shelob9 Or if you would like to specify the TTL, you can change those two lines 
```
LiteSpeed_Cache_API::set_cache_private() ;
[...]

PR #1971, opened 2017-09-25
[NEW] LiteSpeed cache plugin comaptibility

hailite:
Support LiteSpeed Cache plugin with ESI nonce

### DIFF
--- a/classes/core.php
+++ b/classes/core.php
@@ -144,6 +144,11 @@ class Caldera_Forms {
 			add_action( 'wp_ajax_cf_email_save', array( 'Caldera_Forms_Email_Settings', 'save' ) );
 		}
 
+		// Add cache plugin support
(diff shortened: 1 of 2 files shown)


Shelob9:
For #1968

Shelob9:
@hailite This looks good. I like having this conditional logic in place vs a filter.

Before I merge, can you confirm this solves the compatibility issue? I have no way to test.

Besides that and inline doc issues (I'll fix after merge) looks fine to me.

hailite:
Yes, this is from my test code with your plugin and our pre release version, which is on https://github.com/litespeedtech/lscache_wp/tree/master/litespeed-cache

hailite:
Please tell the version this can be added, so we can add your plugin into our compatible list.

If you want, you can add us too.

Thanks.

Shelob9:
@hailite This is now merged and will be included in Caldera Forms 1.5.6.2. Thanks for your help.

hailite:
Gotcha. Thanks.

PR #1986, opened 2017-10-04
QCOSS-113: Check if the processed magic value is same as the default value

agilewarealok:
## Before
If the form has two processors (A & B), one of the processor has magic tags (A) and second processor(B) is using those magic tags to store the values. If some conditions matched Processor A is disabled and only Processor B is working. In this case Processor B is saving the magic tag values as {processor_a:field_name} This should not happen, if value is not present we should save blank values.

## After
Now we will save blank values for invalid magic tag values.

## Solution
We're checking the default value with returned value by bracket magic. If both are same we consider it is an invalid value and make it blank.

### DIFF
--- a/classes/core.php
+++ b/classes/core.php
@@ -757,7 +757,7 @@ class Caldera_Forms {
 				'entry_id' => $entry_id,
 				'field_id' => $field[ 'ID' ],
 				'slug'     => $field[ 'slug' ],
-				'value'    => self::do_clean_magic_tags( $entry )
+				'value'    => self::do_magic_tags( $entry )
 			);
 
 			// named key kets .key to slug
@@ -786,23 +786,6 @@ class Caldera_Forms {
 		}
 
 	}
-
-	/**
-	 * QCOSS-113: Clean magic tags value
-	 * Check if the processed magic value is same as the default value.
-	 * If yes Make it blank. Required when processor is linked to magic tags and due to some conditions
-	 * they do not have actual value.
-	 */
-	 public static function do_clean_magic_tags($value) {
-		 $magicvalue = self::do_magic_tags( $value );
-		 $magics = Caldera_Forms_Magic_Util::explode_bracket_magic( $value );
-		 if ( ! empty( $magics[ 1 ] ) && $value == $magicvalue) {
-			 $value = "";
-		 } else {
-			 $value = $magicvalue;
-		 }
-		 return $value;
-	 }
 
 	/**
 	 * Save final form data
--- a/classes/core.php
+++ b/classes/core.php
@@ -757,7 +757,7 @@ class Caldera_Forms {
 				'entry_id' => $entry_id,
 				'field_id' => $field[ 'ID' ],
 				'slug'     => $field[ 'slug' ],
-				'value'    => self::do_magic_tags( $entry )
+				'value'    => self::do_clean_magic_tags( $entry )
 			);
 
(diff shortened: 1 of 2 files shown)


classes/core.php
```diff
@@ -823,6 +823,23 @@ public static function save_field_data( $field, $entry_id, $form ) {
 	}
 
 	/**
+	 * QCOSS-113: Clean magic tags value
+	 * Check if the processed magic value is same as the default value.
+	 * If yes Make it blank. Required when processor is linked to magic tags and due to some conditions
```
Shelob9: @agilewarealok I 100% agree with your goal here. If the value was removed by conditional logic, return should be empty value.

But, what if the value should be default? IE it was never changed  from default.

Scenario I'm concerned about - field has default value of 10, I don't change that, so doesn't L834 evaluate true here?

Shelob9:
@agilewarealok I opened #1989 for the bug in question.

classes/core.php
agilewarealok: We've covered that case.
We will first check the field value by `explode_bracket_magic`, which checks if the value consists magic tags value or not. We will only compare new and old values if default value is magic tags value otherwise we return the processed value (Default value in this case)

Shelob9: Ok, that makes sense. I will test with that in mind.

FYI I'm keeping this on 1.5.7 milestone. I'm hoping to get 1.5.6.2 out on Monday and keep it to calculation field improvements. I will merge this once 1.5.6.2 is tagged.

agilewarealok: Thanks @Shelob9 , Appreciate it.

Shelob9:
@New0 RE: This fix not affecting emails, it's because this code parses magic tags for emails: https://github.com/CalderaWP/Caldera-Forms/blob/master/classes/save.php#L241-L251

That code isn't using newer APIs. The correct fix is to replace that old out of date way of doing it with the new way, so that its behaviour is consistent.

New0:
@Shelob9 
- Should we consider emails part ( as it not mentionned in the issue )?
- Should we consider extra content in a field, texts + bracket magic tag ?

The fix proposed only works for bracket magic tags that are used alone in single text fields but not along other content.

Shelob9:
@New0 Please create a new PR (this one is against master branch so we can't merge it) with this code and also please use it as a way to fix the email magic tag parsing, since that's needed to get your fix to work:

https://github.com/CalderaWP/Caldera-Forms/blob/master/classes/save.php#L241-L251

jusfreeman:
@Shelob9 which branch do you want this PR submitted for? develop or other?

New0:
Hello @jusfreeman , thank you for this PR. Please do PR against develop branch.

However, the new do_clean_magic_tags only work in single line text fields only using a bracket magic tag.
 If we have text and a bracket magic tag in the field, the do_clean_magic_tags function won't apply.
In order to apply it to email messages we would need to try and set the unused bracket magic tag to null even when it is used along other content, is that correct ?

agilewarealok:
Hello @New0 
Yes, You right right. Fix which we did earlier was only for single text field and not for mixed text and magic fields. We've updated the `Caldera_Forms_Magic_Doer::do_field_magic` method to replace the value if it is same after processing. So we don't need `do_clean_magic_tags` anymore and we've removed the same.

We've created new PR against develop branch with fix. Closing this issue as it's not required anymore.

PR #1991, opened 2017-10-07
Fix hidden fields not updating in calculations properly

### LINKED ISSUE #1979: Calculation fields not updated on first click - sample form attached
no

## Is This A Feature Request?
no

WordPress Version: 4.8.2PHP Version: 5.6.30MySQL Version: 5.6.32Caldera Forms Version: 1.5.6.1

The issue has been observed on Mac Safari and Chrome as well as Windows Edge and Chrome. Oh, and iPad Safari too. 

yes, tested on brand new WP install with no other plugins added.

values for hidden fields are not properly handled in calculation fields when conditions are met. You must click the checkbox, un-check and then re-check before the values of the hidden fields are properly handled. In the attached sample form, you should be able to check the first and second boxes and immediately see the discount of 10 applied. But it does not. Instead you must check the first and second box, un-check the second box, then re-check it. A summary field is included in the form so that you can observe the values being assigned to the hidden fields.

Expected behavior:
If item 1 is checked, no discount. If items 1 and 2 are checked, 10 dollar discount. If items 1, 2, and 3 are checked, discount is 10 + 20. We should not have to click a box 3 times to get the form to calculate.

Replication:
Built a form that calculates prices and discounts based on selection of a checkbox element. Added sub total, discount and grand total calculated fields. Added hidden fields to contain the discount amounts. created conditions set to 'show' each hidden field based on user selection. Example form is attached. You can easily import this into Caldera and test it yourself. 

absolutely no PHP errors appeared while WP_DEBUG set to true

absolutely no javaScript errors were observed in the developer tools console

[bug-report-export.json.zip](https://github.com/CalderaWP/Caldera-Forms/files/1345399/bug-report-export.json.zip)

Shelob9:
Fixes #1979

I submit as evidence that this bug is fixed that the tests described in this comment pass:
https://github.com/CalderaWP/Caldera-Forms/issues/1979#issuecomment-334901439

I will add other tests for other recent calculation issue first.


Shelob9:
This PR might get merged into work for  #1993

Issue #1993, opened 2017-10-07
Value of number field not respected by calculation if had a default


Shelob9:
Scenario - Create a number field or single line field with a default of 1 and a calculation field that with one part, that number field. You would expect that the calculation field starts at 1, and rises to 2 when you change the number field to 2. 

I have written a test that confirm that this is not a bug in 1.5.6, but is in the current master, and in feature/1979 so I believe this is a regression from a recent fix in 1.5.6.2 milestone. My test shows that it does not, and has a second calculation field + number field with out a default and it works fine.

This test, once it passed will show that the field 

Since this is a regression from a recent fix in 1.5.6.2 milestone, probably #1962 or #1979 I don't want a new feature branch since feature/1979 is still unmerged -- PR #1991 I will make feature/1979-1993 and if it fixes both, merge together, else merging both is going to be a pain.

What's cool is I caught this with the test to prove #1962 was really fixed.

- Last working version 1.5.5

Shelob9:
# Details about test `1993 Calculations created@1.5.5 dev@1.5.6.2`

## Test proves that:

- A number field with a default will contribute that default value to a calculation and changes to that field will be reflected in the calculation.
- A number field without a default will contribute 0 to a calculation and changes to that field will be reflected in the calculation. 
- A third calculation field based on the other two will continue to calculate the right value as the other two change.

## URLS for test
- Ghost inspector https://app.ghostinspector.com/tests/59d96bd330e3b074c50cf523
- Testing site: http://dev-test-cf-ga.pantheonsite.io/1993-calculations-created1-5-5-dev1-5-6-2/

Shelob9:
The commit 9d4543d did not pass its own acceptance tests or the test suite's acceptance tests, specially those for #1962 . See comment here: https://github.com/CalderaWP/Caldera-Forms/pull/1995#issuecomment-334979297 issue #1993 is therefore not fixed.

Shelob9:
Close via
7164eba

Issue #1962, opened 2017-09-19
Radio and checkbox defaults not set in state properly if using calc value


Shelob9:

Shelob9:
Workaround until I come up with a better solution, need to think on this -

```
add_action( 'wp_footer', function(){
	?>
	<script>
		jQuery( document ).ready( function( $ ){

			$( document ).on( 'cf.form.init', function( e, obj ){

				var $fields = $( '#' + obj.idAttr + ' input:checkbox, ' + '#' + obj.idAttr + ' input:radio' );
				if( $fields.length ){

					$fields.each( function (i,field) {
						$( field ).trigger( 'change' );
					});
				}

			});
		});
	</script>

	<?php
});

```

synonymouse:
Guys, selects doesn't work too.

Shelob9:
Close via 	f43c350

Shelob9:
Test https://app.ghostinspector.com/tests/59d828c430e3b074c50ccf39 asserts that the initial value is provided to the calculation from a checkbox, dropdown and radio field and that those calculations work well with other calculation fields.

PR #1995, opened 2017-10-08
Fixes for calculation fields

### LINKED ISSUE #1979: Calculation fields not updated on first click - sample form attached

Shelob9:
Fixes #1979 #1993


Shelob9:
The test for this #1962 [try it for yourself](http://dev-test-cf-ga.pantheonsite.io/1962-calculations-created1-5-5-dev1-5-6-2/) is failing now. The last calculation's field state isn't updating.

I think I need to make it so the calculation fields defaults are set server-side, presuming that code is abstract enough. The current system relies on the calculation fields to run on first load and then have that change a) make the DOM look right and b) initialize state. that's a) wasteful b) bad design. But let me let the dog out and eat and possibly eat and watch TV first.

Shelob9:
I'm merging. All test forms work in browser manually, and automated testing and the emails have the right values.

The amount of infrastructure I built to test this one is amusing to me. Really excited about this one.

Issue #2001, opened 2017-10-12
Role parameter in caldera_forms_entry_viewer shortcode leads to error


New0:
The role parameter seems to always print Sorry you are not allowed to view this and give the response
{"code":"rest_forbidden","message":"Sorry, you are not allowed to do that.","data":{"status":403}}

Shelob9:
Close via #2003

PR #2003, opened 2017-10-12
Fixed issue #2001

New0:
The role parameter seems to always print Sorry you are not allowed to view this and give the response
{"code":"rest_forbidden","message":"Sorry, you are not allowed to do that.","data":{"status":403}}


Shelob9:
@New0 Please add a test for this. Will require an entry viewer test suite be added.

Also, we will need to run the test while logged in and not logged in. I am working on how to make that possible.

Shelob9:
Fixes #2001

PR #2083, opened 2017-11-16
Fixes BCC visible to Email Recipients #1504

### LINKED ISSUE #1504: BCC visible to Email Recipients
WordPress Version: 4.7.3
PHP Version: 7.1.3
MySQL Version: 5.6.28
Caldera Forms Version: 1.5.0.7

Yes.

Consider three email addresses, [A],[B], and [C].
I set up a contact form to email to recipient [A], and BCC to [B, C]. On submission, Only [A] and [B] received the email, and they both were listed as normal recipients (not BCC) in both inboxes. [C] did not receive the email at all. I've tried excluding [C], but that didn't make [B] behave like a real BCC recipient.

I'm using SendGrid, and can confirm that [C] was never even attempted to be processed/blocked/bounced at SendGrid, so it is not an email client issue. I have tested email [C] with Caldera Forms and it has been able to receive submissions. 

None since this issue.

None.

New0:
Fixes #1504 BCC visible to Email Recipients with @nathanhere suggestion

### DIFF
--- a/classes/core.php
+++ b/classes/core.php
@@ -1085,14 +1085,14 @@ class Caldera_Forms {
 		}
 
 
-        $email_message['bcc'] = false;
-        if (  ! empty( $config[ 'bcc' ] ) ) {
-            $email_message['bcc']       = Caldera_Forms::do_magic_tags( $config[ 'bcc' ] );
+		$email_message['cc'] = false;
+        if (  ! empty( $config[ 'cc' ] ) ) {
+            $email_message['cc']       = Caldera_Forms::do_magic_tags( $config[ 'cc' ] );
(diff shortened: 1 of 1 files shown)


Shelob9:
@new0 this looks good. I will test tomorrow.

Can you confirm you tested with BCC set with field magic tag and hardcoded value with single and multiple BCCs in each configuration already?

New0:
Yes @Shelob9 , the Bcc addresses are hidden, I testest with hard-coded and magic tags values with one or up to three addresses, in both the Bcc and email address fields.

I just noted that the Bcc field is not used by the pro app

PR #2088, opened 2017-11-18
Fixed some WYSIWYG issues.

0xLBF:
When editing a post with the WYSIWYG field a bunch of unnecessary whitespace was inserted and the html code got escape twice, leading to the html code being displayed instead of the content.


Shelob9:
@0xLBF Thanks for your contribution and for being willing to resubmit (I hate asking for that, but I have to.)

PR #2091, opened 2017-11-20
Remove uneeded role attributes

### LINKED ISSUE #1750: HTML validation issues
Caldera Forms v1.5.2.1 on WP v4.8 

Could you please fix html validation issues, I have attached a patch for review.  Here are the issues.

- "field" is not a valid value for the "role" attribute.
- "role" attribute with the value of "form" is not needed on a "form" element.
- When a label is hidden for a particular field, "labelledby" attribute is mislabeled. 

[caldera-forms.txt](https://github.com/CalderaWP/Caldera-Forms/files/1149285/caldera-forms.txt)

Shelob9:
Fixes #1750

### DIFF
--- a/classes/core.php
+++ b/classes/core.php
@@ -3876,7 +3876,7 @@ class Caldera_Forms {
 			"field"             => $field,
 			"id"                => $field[ 'ID' ],//'fld_' . $field['slug'],
 			"name"              => $field[ 'ID' ],//$field['slug'],
-			"wrapper_before"    => "<div role=\"field\" data-field-wrapper=\"" . $field[ 'ID' ] . "\" class=\"" . $field_wrapper_class . "\" id=\"" . $field_id_attr . "-wrap\">\r\n",
+			"wrapper_before"    => "<div data-field-wrapper=\"" . $field[ 'ID' ] . "\" class=\"" . $field_wrapper_class . "\" id=\"" . $field_id_attr . "-wrap\">\r\n",
 			"field_before"      => "<div class=\"" . $field_input_class . "\">\r\n",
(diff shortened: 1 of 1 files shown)


classes/core.php
```diff
@@ -4119,7 +4119,6 @@ static public function render_form( $atts, $entry_id = null, $shortcode = null )
 		$form_attributes = array(
 			'method'	=>	'POST',
 			'enctype'	=>	'multipart/form-data',
-			'role'		=>	'form',
```
Shelob9: For #1750

New0:
@Shelob9 Are we sure we want to remove the role attribute from form element ? 
It is not necessary but tools like JAWS use it. It makes the form to be interpreted as a region and some people are used to navigate with the R shortcut

Shelob9:
I consulted with @bamadesigner who knows better than I do here. She says: `https://www.w3.org/TR/wai-aria-practices/examples/landmarks/form.html
"If a form element has an accessible name it is considered form landmark."`

Since we do not have a label for forms, I think we should not output a a role attribute, but introduce the filter so that Easy Pods and Easy Queries can set `role="search"`

bamadesigner:
Amanda is correct. Per that Stack Overflow question, you should always follow the standards and not what 1 specific piece of software does. <form>s do not need `role="form".`

Also, don't forget the aria-label for the form so ARIA does know what the form is for and the form has an accessible name.

Shelob9:
Ok, so I'm sold on removing role attribute. I actually think we can get the role=search in for Easy Pods/ Easy Queries search forms without a new filter. 

@bamadesigner Your point about aria-label I agree with, but out of scope of this PR. I created  a new issue #2093

New0:
Ok me too @Shelob9 , @bamadesigner convinces me, we better follow the standards

Shelob9:
I have updated the docs for the caldera_forms_render_form_attributes filter with an example of how to set the role attribute to search.

https://calderaforms.com/doc/caldera_forms_render_form_attributes/

Issue #2082, opened 2017-11-16
Default value of select fields only handled with calculation value


New0:
Reproducible issue at :
- https://secure.helpscout.net/conversation/430142856/5136/

When a select field value is involved in a calculation, it always need to be set manually. Otherwise a default value that remain unchanged won't be taken into account by the calculation

New0:
Default value of select fields only handled with calculation value when manually set select field with empty calculation field and a value field work

Shelob9:
@New0 I'm assigning this to you.

Shelob9:
Related #2090

Shelob9:
Also probably related  ##2076

Shelob9:
Reassigning to me

Shelob9:
Test "2082 Calculations created@1.5.7.1 dev@1.5.8" now exists for this issue.

Shelob9:
b0ba255 fixes #2090 This issue is not yet fixed.

Shelob9:
Test "2082 Calculations created@1.5.7.1 dev@1.5.8" covers this.

Shelob9:
Close via  #2143

Issue #2063, opened 2017-11-09
Calculation not working with Summary


101010f:
## Do You Need Immediate Help?
No

## Is This A Feature Request?
No

## What Version Of Caldera Forms, WordPress and PHP Are You Using?
1.5.7 / WordPress 4.8.3

## Does Your Issue Persist When You Disable All Other Plugins and Switch To The Default Theme?
Yes

## What Is The Unexpected Behaviour?
Calculation fields are always 0 if a summary field is added to the form
![capture](https://user-images.githubusercontent.com/20680030/32606607-9fe62b8c-c55e-11e7-851b-7388a4d8999c.PNG)

## What PHP Errors Have You Logged While Reproducing This Bug?
None

## What JavaScript Errors Have You Seen While Reproducing This Bug?
None

Shelob9:
`1979 CALCULATIONS CREATED@1.5.5 DEV@1.5.6.2` Is the test that we were relying on for this. It's totally working too -- http://dev-cf-testing.pantheonsite.io/1979-calculations-created1-5-5-dev1-5-6-2/

That form has lots of conditionals -- exposes how they were breaking summaries and calculations in some cases before. I made a new test with no conditionals -- `2063 Calculations create@1.6.2.2 dev@1.5.7.1` ISSUE REPRODUCED - http://dev-cf-testing.pantheonsite.io/2063-caluclations-create1-6-2-2-dev1-5-7-1/

Shelob9:
This doesn't work in 1.5.6.2 or 1.5.5 or 1.5.4 http://r-154-cf-testing.pantheonsite.io/2063-caluclations-create1-6-2-2-dev1-5-7-1/

Shelob9:
Test 5a0486d1d5628e428b3399f1 is for this. I faked the baseline, failing on 1.5.7, which is good.

Shelob9:
Test 5a0486d1d5628e428b3399f1 is also for this, passes on 1.5.6.2, failing on 1.5.7, which is good.

Shelob9:
This is actually the same issue as #2082 and is fixed and tests are passing after #2143

PR #2098, opened 2017-11-26
Validate advanced file fields on multi-page forms properly

### LINKED ISSUE #2021: Advanced file field required option doesn't block users to click and go to next page
WordPress Version: 4.8.2
PHP Version: 7.0.20-2~ubuntu14.04.1+deb.sury.org+1
MySQL Version: 5.5.5
Caldera Forms Version: 1.5.6.1

On a multi page form, the advanced fiel field required option doesn't block users to click and go to next page, when I get to the submit button the form is not sent but no errors are thrown ( no console error or debug.log error)

Shelob9:
Fixes  #2021

This fix looks good. I need to create an acceptance test for single page and multi page forms. Also need to make sure this didn't break any other type of validation, before I'm comfortable merging. Will work on that after lunch.

Also I would like to use [cf-result-diff-plugin](https://gitlab.com/caldera-labs/cf-result-diff-plugin) to prove the bug and fix, in terms of email content and what is saved in the database, which it almost is capable of doing.

@New0 I would like your review as well please.


New0:
It worked in my local tests, the next page action is blocked when no file is uploaded in a required advanced file upload field. And it is still submitting the files when they are uploaded.

Shelob9:
Test for single page validation (not really the issue here, but we want to ensure this fix doesn't break that: 2021.1 Adv File created@1.5.7.1 dev@1.5.8

Test for multi-page validation 2021 Adv File created@1.5.7.1 dev@1.5.8

Both of which pass on this branch.

Shelob9:
I'm going to merge this even though test "2063 Calculations create@x dev@1.5.7.1" is failing. This is because #2082 is probably a new symptom of #2063 but only partially so, so both test for 2082 and 2063 should cover that before release.

Issue #2107, opened 2017-11-28
Credit card fields validate on wrong page of mutli-page forms


Shelob9:

Shelob9:
The form in the test sheet 2107 Credit Card created@? dev@1.5.8 reproduces this.

blunket:
Glad to see this is being worked on! Feel free to close #2095 now (I opened that one about this same issue).

Shelob9:
@blunket Yep, I'm just describing it different here. Removing that function isn't a long-term solution, as it's supposed to disable advance if field is invalid __and it's on the current page__.

Shelob9:
Note to future self, and others like me: you can determine what page a field is on, like this:
https://gist.github.com/Shelob9/ee38f435b9739f7821ab542af551be64

Shelob9:
Should be fixed via 
7f2f6b2

Test "2107 Credit Card created@? dev@1.5.8" is defined in sheet, but not yet recored, will record tomorrow.

Shelob9:
Test "2107 Credit Card created@? dev@1.5.8" is now created and passing on this PR.

Shelob9:
Close via #2139

PR #2139, opened 2017-12-13
Fix credit card field validation 

Shelob9:
See notes about testing: https://github.com/CalderaWP/Caldera-Forms/issues/2107#issuecomment-351261305

For #2107


Shelob9:
Merging even though test for #2063 is failing as that's a separate issue.

Issue #2090, opened 2017-11-19
No empty select option if 0 is default.


jpgoem:
## Do You Need Immediate Help?
* No

## Is This A Feature Request?
* No

## What Version Of Caldera Forms, WordPress and PHP Are You Using?
WordPress Version: 4.9 PHP Version: 7.1.10 MySQL Version: 5.5.5 Caldera Forms Version: 1.5.7.1

## Does Your Issue Persist When You Disable All Other Plugins and Switch To The Default Theme?
Yes

## What Is The Unexpected Behaviour?
If a dropdown select field has **0** as one of its options, **and** that value is the default, then an additional and unexpected blank (or null) option is listed in the dropdown list.

This appeared first **in CF 1.5.1** and is still there in 1.5.7.1.   Up to CF 1.5.0.10 there was no blank (or null) option in the dropdown list.

screenshot in CF up to 1.5.0.10 : 
![image](https://user-images.githubusercontent.com/31822130/32990241-721cd0fa-cd26-11e7-991e-7ab194f2830e.png)

screenshot in CF from 1.5.1 and later (tested in 1.5.7.1) : 
![image](https://user-images.githubusercontent.com/31822130/32990264-c7fc7c14-cd26-11e7-9a9c-cc19edda6220.png)

Here is a sample Json export : 
[dropdown-and-default-zero-export.zip](https://github.com/CalderaWP/Caldera-Forms/files/1485504/dropdown-and-default-zero-export.zip)

## What PHP Errors Have You Logged While Reproducing This Bug?
None

## What JavaScript Errors Have You Seen While Reproducing This Bug?
None

Shelob9:
Need to make sure not to break fixes in #2013 when fixing this.

Also, let's make sure #2082 gets fixed without complicating this.

Shelob9:
Test in sheet "2082 Calculations created@1.5.7.1 dev@1.5.8" is based on form export form 	@jpgoem and reproduces this.

Shelob9:
Close via b0ba255ac7ba3943e30f20bbfa399da6fcb7d046

PR #2142, opened 2017-12-13
don't put empty option if default is 0 

Shelob9:
For #2090


fields/dropdown/field.php
```diff
@@ -33,7 +33,7 @@
 			}
```
Shelob9: For #2090

Shelob9:
Test "2090 Select created@1.5.7.1 dev@1.5.8" pasess. No new fails

Issue #2147, opened 2017-12-13
Buttons in admin editor header are inaccessible


bamadesigner:
WordPress Version: 4.9.1PHP Version: 5.6.31MySQL Version: 5.5.5Caldera Forms Version: 1.5.7.1WP_DEBUG: 1

Right now the "buttons" in the editor header are inaccessible by keyboard. They are <a> without href attributes which makes them untabbable.

It's an easy fix. Changing the "buttons" to be actual buttons will do the trick. I have a PR coming.

bamadesigner:
I resubmitted the PR: https://github.com/CalderaWP/Caldera-Forms/pull/2152

Sorry about that. I was so close to following your contributing rules!

Shelob9:
Close via b3bbc96

PR #2148, opened 2017-12-14
Changing admin editor actions to buttons

bamadesigner:
Changing the "buttons" in the admin header to be actual buttons so they will be tabbable and therefore accessible. Right now they are <a> without href attributes which makes them untabbable and therefore inaccessible by keyboard.

For https://github.com/CalderaWP/Caldera-Forms/issues/2147


Shelob9:
@bamadesigner Sorry I have to do this, but I have to close this as its against the wrong branch. Can you resubmit against develop branch please? It would be a mess to merge against master.

https://github.com/CalderaWP/Caldera-Forms/compare/develop...bamadesigner:feature/2147

PR #2149, opened 2017-12-14
Fix bugs in validating token on re-entering process via url

ChrisFlannagan:
Fix an issue when trying to continue with an entry submit using the cf_et and cf_ee query string parameters.  It was using the token for entry id in validation and the is_wp_error check was returning an error if there was not one instead of the opposite.


Shelob9:
@ChrisFlannagan I merged your code to a different branch via #2158 I want to do a little more clean up on the feature.

PR #2152, opened 2017-12-14
Changing admin editor actions to buttons for accessibility

bamadesigner:
Changing the "buttons" in the admin header to be actual buttons so they will be tabbable and therefore accessible. Right now they are without href attributes which makes them untabbable and therefore inaccessible by keyboard.

For https://github.com/CalderaWP/Caldera-Forms/issues/2147

@Shelob9 Sorry about the wrong branch PR. I was so close to following your contribution rules. :)


Shelob9:
Thanks Rachel.

PR #2155, opened 2017-12-15
Hide breadcrumb after submission, for multi paged forms

### LINKED ISSUE #1702: Hide breadcrumbs after submission
## Is This A Feature Request?
Yes

When a multi-page is submitted the breadcrumbs stay visible even when the form setting 'Hide Form' has been enabled.
I think it would make more sense if the breadcrumbs would also be hidden after submission.

image

New0:
Fix #1702 enhancement


Shelob9:
This works for me when doing manual testing. We need to record test first.

New0:
Tested by https://app.ghostinspector.com/tests/5a39863b3be6b23ca8d8be17

Issue #2163, opened 2017-12-20
Merge Caldera Forms Pro API client into CF


Shelob9:
For now, we will keep to separate build processes. Once we figure out GutenPlan, we will decide on one build process to rule them all.

Must remove the Github updater and in build process, the source files for Vue app.

Shelob9:
Fixed for now. New issues caused by this (not bugs but redundancies)

- #2168
- #2166 
- #2167

Shelob9:
Close via 
8e5ca90

PR #2165, opened 2017-12-20
Merge cf pro into core

Shelob9:
#2163


Shelob9:
I'm merging this now. For final QA we will leave CF Pro client installed on test site.

PR #2187, opened 2018-01-03
Stop notice set up if not for current form

nigelbyrne82:
Additional condition to prevent notices being built for other forms by comparing the transient ($prev_post['data']['_cf_frm_id']) form id to the currently processed form ($form['ID'])


Shelob9:
@nigelbyrne82 Thanks for your pull request. I appreciate your contribution. The change looks good to me.

@New0 Please preform the following QA when the task for reviewing your PR comes up in Asana:

* TL;DR We need to make sure this patch doesn't break advanced file fields, or the PayPal Express add-on.

- Open an issue describing what this fixes. Something like "With two forms on a page, errors stored in transient show on both forms".
- Figure out how to reproduce this. I think to reproduce this you'd need two forms, different IDs, on the same page, and submit one with an error that would trigger a server side validation error. (Use filter `'caldera_forms_validate_field_' . $field[ 'type' ]`  it's in `Caldera_Forms::process_submission()` to trigger errors by returning a `WP_Error` object on that filter)
- Once you can reproduce it against develop, check out this branch.


Transient storage is used to carry data between different HTTP requests that make up 1 distinct form submissions. Traditionally `$_SESSION` would solve this problem -- carrying data between requests as one distinct session -- because shitty shared hosts and Docker (LOL, right?) make session storage suspect, so we use the database or persistent object cache instead. We wrap the WordPress transients API using a class called `Caldera_Forms_Transient`.

Keep in mind that, form submissions that require more than one HTTP request happen when a submission makes use of an advanced file field, which is sent in one or more separate asynchorous HTTP requests and then once they are all done -- technically speaking the promise is resolved successfully -- then the rest of the form data submits, or someone re-submits the form after a server-side validation failed or a processor returned an error. Also, PayPal express, form submits, transient is stored, redirect to paypal, commerce happens, redirect back to WordPress, Caldera Forms picks up the submission using the data in the transient.

We can't, until next iteration of acceptance testing, test two forms on a page with automation. But you can run these tests manually. And we have automated acceptance tests for file fields.

PR #2205, opened 2018-01-06
Fix json_decode expects parameter 1 to be a string admin.php L963

### LINKED ISSUE #1996: json_decode produces warning with code using get_entries for certain types of entries
* WordPress Version: 4.8.2
* PHP Version: 7.0.24
* MySQL Version: 5.6.36
* Caldera Forms Version: 1.5.6.1
* WP_DEBUG: false

Yes

Using [this code](https://calderaforms.com/doc/get-entries-from-a-form/) in Docs produces a PHP warning:
`json_decode() expects parameter 1 to be string` on this line: 
https://github.com/CalderaWP/Caldera-Forms/blob/master/classes/admin.php#L966

It doesn't happen for all forms/entries. It works fine for some and not for others. The only difference I can tell is that the forms it's not working for have entries that are showing some custom fields in the entries table whereas others are just showing the default fields in the entries table.

## Potential Fix
Switching the line above from `json_decode` to `maybe_unserialize` seems to resolve the issue but would need further testing.

New0:
Closes #1996 Caldera_Forms_Admin::get_entries returned a notice for json_decode that actually caused issues.

### DIFF
--- a/classes/admin.php
+++ b/classes/admin.php
@@ -960,11 +960,11 @@ class Caldera_Forms_Admin {
 							$field = Caldera_Forms_Field_Util::get_field(  $row->slug, $form, true );
 
 							// maybe json?
-                                if ( is_string($row->value) ) {
-                                    $is_json = json_decode( $row->value, ARRAY_A );
-                                } else if ( is_array($row->value) ) { //Process all checkbox values
-                                    $is_json = $row->value;
-                                }
+                                            if ( is_string($row->value) ) {
+                                                $is_json = json_decode( $row->value, ARRAY_A );
+                                            } else if ( is_array($row->value) ) { //Process all checkbox values
+                                                $is_json = $row->value;
+                                            }
 
 							if ( ! empty( $is_json ) ) {
 								$row->value = $is_json;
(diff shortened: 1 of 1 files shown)


classes/admin.php
```diff
@@ -960,7 +960,12 @@ public static function get_entries( $form_or_id, $page = 1, $perpage = 20, $stat
 							$field = Caldera_Forms_Field_Util::get_field(  $row->slug, $form, true );
 
 							// maybe json?
-							$is_json = json_decode( $row->value, ARRAY_A );
+                            if ( is_string($row->value) ) {
```
Shelob9: @New0  can you fix these indents please ? 

related #2206

Shelob9:
@New0 This looks good.

Please do the following QA:

1) Check this PR out on the test site and run all tests. I can't see how this would cause any of them to fail since they are all front-end, but sometimes code in `Caldera_Forms_Admin` is used for something front-end. Those "sometimes" get us.
2) Create a form with 3 checkbox fields. Submit with field 1 having checked one option and field 2 having checked multiple options and field 3 having no options correct. Observe how this data is represented in the entry viewer, email, auto-responder and database.

Issue #2108, opened 2017-11-29
Provide tools for GDPR compliance


Shelob9:
# This is the overview/ roundup issue see [this page](https://calderaforms.com/gdpr) 

Given the amount of data collected by Caldera Forms, we must provide the right tools for GPDR compliance by the 28th of May 2018. I think this is a good thing as our users are sitting on a pile of private data and a lot of times, deleting it is great solution to avoid privacy concerns and trim database size.

Everything we do has to be opt-in by site admin. I think most sites should enable this compliance, but I don't think we should make that call for them and some may have other strategies for compliance.

We will use the standard proposed here: https://github.com/GDPRWP/standard

What we need is the following:

- New method to get all entries where user ID is a given ID.
- New method to get all entires where given email is the value of a given field.
- New method to delete entires matching those results.

- Admin-facing user interface screen for GPDR (maybe we call this "User Privacy")
-- Setting to enable discovery and request deletion
-- Setting to restrict look up to logged in users. This is the default, we can't just let anyone enter email addresses into the site 
-- Setting for 

- New form processor for GPDR
-- Search by email - return results -- return results with delete option
-- Search by current logged in user -- return results with delete option

- Template for a GPDR Form
-- Look up fields
-- Privacy policy display field

UniversalDesignz:
Hi Josh,

You guys are doing an awesome job. I am looking forward to the new updates with GPDR compliance implemented in Caldera forms? It would be nice to have an option that would allow to choose auto deletion after scheduled day(s) | month(s) | Year(s) or indefinite time.

Thank you.
Kindest Regards
Jay

Shelob9:
Update on our work on GDPR compliance:

We need two basic things:
1. A query builder tool that can select and delete the kind of data needed to comply with GDPR.
2. A system with an interface and stuff for configuring, managing and handling these requests and the privacy statement and...

🌋[Item one is now ready to use.](https://github.com/CalderaWP/caldera-forms-query). It is a composer package, not a plugin. It is a developer tool that is also going to be useful for improving entry viewing, editing, display and exporting to CSV.

[...]

PR #2273, opened 2018-02-01
QCOSS-113: Check if the processed magic value is same as the default value

agilewarealok:
## Before
If the form has two processors (A & B), one of the processor has magic tags (A) and second processor(B) is using those magic tags to store the values. If some conditions matched Processor A is disabled and only Processor B is working. In this case Processor B is saving the magic tag values as {processor_a:field_name} This should not happen, if value is not present we should save blank values.

## After
Now we will save blank values for invalid magic tag values.

## Solution
We're checking the default value with returned value by bracket magic. If both are same we consider it is an invalid value and make it blank.

Same PR: #1986  against development branch.

### DIFF
--- a/classes/core.php
+++ b/classes/core.php
@@ -757,7 +757,7 @@ class Caldera_Forms {
 				'entry_id' => $entry_id,
 				'field_id' => $field[ 'ID' ],
 				'slug'     => $field[ 'slug' ],
-				'value'    => self::do_magic_tags( $entry )
+				'value'    => self::do_clean_magic_tags( $entry )
 			);
 
 			// named key kets .key to slug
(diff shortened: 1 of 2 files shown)


agilewarealok:
@New0 @Shelob9 
What you think about the new fix ? We've updated `do_field_magic` to do the magic.

jusfreeman:
@New0 @Shelob9 just a friendly poke to see if this PR will be reviewed, accepted?

Shelob9:
@jusfreeman This PR is very interesting, I'd love to have it merged. It also creates a lot of risk for unintended side-effects. It was not a high enough priority issue to place on the 1.6.0 milestone, and assign developer time to creating automated tests for. We will see if we can invest that time once #2108 is resolved.

jusfreeman:
@Shelob9 Any chance of reviewing this PR, been a while.

Shelob9:
@jusfreeman I think we should review this. It looks good.

@New0 once we get all 1.8.0 issues merged, please review this. 

Here are my concerns:

- This has no tests. Not sure what would be the best way to test this, but getting better coverage on magic tags, which should be doable, in isolation, by injecting mock values directly to the `Caldera_Forms_Magic_Doer::do_magic()` method.
- I'm struggling to think of other side effects that we could run into here, and I can't, which is curious.

jusfreeman:
@Shelob9 sounds good mate - Xmas is coming so there's also that to aim for! 👍

Issue #2275, opened 2018-02-01
CC and BCC for auto-responder are not passed to cf pro properly


Shelob9:
I need to look at git history for the cc/bcc feature and look at what went wrong here. This is why I want to focus testing on interoperability between CF and CF Pro.

Shelob9:
Should have been closed via #2308 and
3f0049f

PR #2276, opened 2018-02-01
Set bcc/cc properly for caldera forms pro API requests

Shelob9:
#2275


New0:
@Shelob9 The PR works, however I just noticed that multiple recipients don't work in main recipients field with ot without Pro app activated. Do you want to fix that here or should I open a new issue ?

Shelob9:
New issue and I'm pretty sure that's a new feature. I don't think that ever
worked, but it could.

New0:
Sets correctly multiple recipient in the auto responder Cc and Bcc fields inwhen Pro enhanced delivery is activated.

New0:
I will merge feature/2278 into feature/2275 and push a new PR

PR #2280, opened 2018-02-02
multiple recipients on auto-responder feature

### LINKED ISSUE #2278: Auto responder main recipients field not working with multiple email addresses
When using multiple email addresses in the main recipients field of auto responder processor, they return attached like --  address1@email.testaddress2@email.test --

Shelob9:
Fixes #2278


Shelob9:
As-is this will only fix the bug when CF Pro _is not active_. Once this is merged, then we can merge #2276 which continues this fix to CF Pro.

New0:
I will merge feature/2278 into feature/2275 and push a new PR

Issue #2301, opened 2018-02-09
Condition that hides a checkbox field does not rebind checkbox options value to a calculation after unhiding


New0:
When a checkbox field is unhidden by a condition and there are options already in checked status, the values of these options don't get rebind to calculation field.

Test in creation l40

Shelob9:
Close via a98484fb2fc02a4c6298742d8f8923fe61b97ca0 #2302

PR #2302, opened 2018-02-09
Rebind checkbox options value when the checkbox field is unhidden

New0:
Fix for #2301 
When a checkbox field is unhidden, the checkbox options that are already in checked status get rebind to calculation field


New0:
Test Line 40 https://docs.google.com/spreadsheets/d/1b2bIuIb8fKZ4NoC7UHV4hoEmw8W0Levkw2h_tJwDryM/edit?usp=sharing

Issue #2293, opened 2018-02-07
From/reply-to settings are not intuitive #71


Shelob9:
This was: https://github.com/CalderaWP/caldera-forms-pro/issues/71

How it works is:

*From setting in mailer and auto-responder becomes reply-to in message sent by app.
Reply-to will always be norely@whatever... this is super important to avoid having messages marked as spam. We may offer domain white labels later, but only once we have a good way to verify domains, since that's even more important for not being marked as spam.
*The problem is this isn't clear. We need to change the field labels in mailer and auto-responder when CF Pro is in use for the form

@New0 Please check if CF Pro is active (see this new function https://github.com/CalderaWP/Caldera-Forms/pull/2289/files#diff-f81f3dff36bf04b7a070d642c858e883R225 added via #2289) 

and 

If enhanced delivery is active for this form.

	    `$settings = \calderawp\calderaforms\pro\container::get_instance()->get_settings()->get_form( 'cfq11' );`

BTW In the mailer panel, I think `$element` is the current form config. That's preferable to doing ` `$settings = \calderawp\calderaforms\pro\container::get_instance()->get_settings()->get_form( esc_attr( $_GET[ 'edit' ]);`

Shelob9:
Close via 7f57b9b #2304

Shelob9:
I am re-oppening as I am getting these errors in the admin against `develop` branch after merge 
```
Fatal error: Uncaught Error: Call to undefined function caldera_forms_pro_is_active() in /app/wp-content/plugins/caldera-forms/ui/panels/emailer.php on line 209

Error: Call to undefined function caldera_forms_pro_is_active() in /app/wp-content/plugins/caldera-forms/ui/panels/emailer.php on line 209
```

Shelob9:
Ignore me, that was git fail on my part.

Shelob9:
https://xkcd.com/1597/

PR #2304, opened 2018-02-12
Adapt text for Reply-to option depending on Pro delivery status

New0:
#2293 If Pro delivery is enhanced for the form :
- Send from email is transformed in Reply-to and doesn't mention magic tags.
- From Email address doesn't Show.


ui/panels/emailer.php
```diff
@@ -195,20 +205,90 @@
 	</div>
 </div>
 
+<?php //Set Different From email and Reply-to text depending on Pro delivery status of the form
+    $enhanced_delivery = \calderawp\calderaforms\pro\container::get_instance()->get_settings()->get_enhanced_delivery();
```
Shelob9: This has to go after conditional, `\calderawp\calderaforms\pro\container::get_instance()` is not defiend if that fails

Issue #2279, opened 2018-02-02
Allow multiple email addresses in Conditional recipients processor


New0:
Allow the use of a coma separated list of email addresses to be used as recipients in the conditional recipients processor recipients field.

Shelob9:
@New0 Can you see if PR #2280 makes this a non-issue? It's possible that now that auto-responders support multiple recipients, conditional recipients will to.

PR #2313, opened 2018-02-16
Add multiple recipients feature to conditional recipients processors

New0:
#2279 Allow comma separated lists of emails to be used in the reipients field of the conditional recipients processor

### DIFF
--- a/processors/classes/conditional_recipient.php
+++ b/processors/classes/conditional_recipient.php
@@ -29,15 +29,22 @@ class Caldera_Forms_Processor_Conditional_Recipient  {
 	 */
 	public static function post_processor( $config  ){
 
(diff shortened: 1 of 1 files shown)


Shelob9:
Looks good, @robwilde please test manually. @New0  Please provide testing instructions.

New0:
To test 
- Add a conditional recipient processor to a form
- set the condition in the processor

All tests run twice, with Pro activated and deactivated.
First test:
- set a list of comma seperated email addresses in the recipients field of the processor
- in the form, match the condition and submit the form

Second test:
- set one email address in the recipients field of the processor
- in the form, match the condition and submit the form

Run one last test with that fails the condition.

Issue #2318, opened 2018-02-22
Warning illegal offset type thrown when custom field processor is used by the Form


New0:
PHP Warning: Illegal offset type in isset or empty in /caldera-forms/classes/magic/doer.php on line 501 / 518 / 522

An array is passed as the key of an array in the magic_tag_meta_prepare() function.

Shelob9:
Maybe related: #2295 
https://github.com/CalderaWP/Caldera-Forms/issues/2295#issuecomment-363984893 ?

Shelob9:
Close via bc4c6e8   #2319

Issue #2295, opened 2018-02-08
File field uploads with CF Pro error


Shelob9:
If a file field is in form -- I don't think it's every configuration, I have to figure that out still, when CF Pro API client creates an instance of a \Caldera_Forms_Entry, $entry_id is null, so nothing works.

https://github.com/CalderaWP/Caldera-Forms/blob/master/includes/cf-pro-client/classes/api/message.php#L190

Shelob9:
[08-Feb-2018 02:23:32 UTC] PHP Fatal error:  Uncaught Error: Call to a member function to_array() on null in /app/wp-content/plugins/caldera-forms/includes/cf-pro-client/classes/api/message.php:190
Stack trace:
#0 /app/wp-content/plugins/caldera-forms/includes/cf-pro-client/classes/send.php(82): calderawp\calderaforms\pro\api\message->add_entry_data(NULL, Array)
#1 /app/wp-content/plugins/caldera-forms/includes/cf-pro-client/classes/hooks.php(100): calderawp\calderaforms\pro\send::main_mailer(Array, NULL, 'CF5a592725305bf', true)
#2 /app/wp/wp-includes/class-wp-hook.php(286): calderawp\calderaforms\pro\hooks->mailer(Array, Array, Array, NULL)
#3 /app/wp/wp-includes/plugin.php(203): WP_Hook->apply_filters(Array, Array)
#4 /app/wp-content/plugins/caldera-forms/classes/save.php(421): apply_filters('caldera_forms_m...', Array, Array, Array, NULL)
#5 /app/wp-content/plugins/caldera-forms/classes/core.php(843): Caldera_Forms_Save_Final::do_mailer(Array, NULL)
#6 /app/wp/wp-includes/class-wp-hook.php(288): Caldera_Forms->save_final_form(Array)
#7 /app/wp/wp-in in /app/wp-content/plugins/caldera-forms/includes/cf-pro-client/classes/api/message.php on line 190

Shelob9:
That error is fixed in PR #2297 which isn't merged yet, but should get merged into this fix as well. But that doesn't fix actual issue that for some fucking reason $entry_id is null. I must know why but also I'm tired AF.

Shelob9:
Possibly related error on user site `PHP Notice:  Undefined index: default in`~/plugins/caldera-forms-pro/includes/calcdefaults.php on line 50`

Shelob9:
PR #2343  is for this.

[...]

PR #2319, opened 2018-02-22
Check if $entry_id is an array or object in magic_tag_meta_prepare( $entry_id ) and return if true

New0:
Prevent wrong value as the key of an array when custom fields processor is used

#2318 When custom fields processor is used, an array is used as the key of an array. 
$entry_id is an array in magic_tag_meta_prepare( $entry_id ) when custom fields is used.

### DIFF
--- a/classes/magic/doer.php
+++ b/classes/magic/doer.php
@@ -497,7 +497,7 @@ class Caldera_Forms_Magic_Doer {
 			self::$entry_details = array();
 		}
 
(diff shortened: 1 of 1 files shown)


Shelob9:
@New0 Why is it an array?


Check how `cf_handle_file_upload()` is used during advanced file uploads. Also, look in Caldera_Forms class around L3480

Shelob9:
This is probably fixed if  #2295 is fixed.

PR #2287, opened 2018-02-06
initial work for checkbox field - multiple default options support

diegoliv:
@Shelob9 this is a draft for allowing support of multiple default options on the checkout field using the `caldera_forms_render_get_field` filter. On my tests it worked fine without errors. The only thing that I couldn't make it work so far is the calculation fields. When multiple checkboxes are set as default, a calculation field for this checkbox stays at 0. Any tips on where can I check to update the js function that calculates these values?

Issue #2167, opened 2017-12-20
Move all build tooling into one system


Shelob9:
Currently we have grunt for release builds and JS/CSS builds. And webpack for the block build, and webpack for the cf-pro client build.

Either the block build or the webpack build should become the one true build process.

I'm putting this on 1.5.8 but may punt as I have to finish thinking through blocks first.

PR #2322, opened 2018-02-23
Checkbox field multiple default

diegoliv:
New PR to match a new branch against the `develop` branch. Supersedes #2287. Also resolves the issue with calculation fields. @Shelob9 ready for you to check!


Shelob9:
I opened #2325 for this.

@diegoliv This looks great, thanks. 

@New0 Can you test this manually please? Use something like this to set two defaults:

```
    add_filter( 'caldera_forms_render_field', function( $field, $form ){
        if( 'fld111' === $field[ 'ID' ] ){
            $field[ 'config' ][ 'default' ] = array(
                1,2
            );
        }
    
        return $field;
    }, 10, 2 );

```

diegoliv:
@Shelob9 awesome! Next step is updating the UI to allow this same functionality without the use of the filter. I'll be sending another PR for that very soon!

Shelob9:
@diegoliv Awesome. I created #2327 for the UI.

My goal is to merge this PR and include the feature in 1.6, but I don't want to add the UI right now, as we're trying to get 1.6 out as soon as we can. 

I put @2327 on the 1.6.1 milestone. If you have a simple PR for the UI, I'll take a look. I'm tempted to use this as my excuse to replace the options controls entirely as the UX needs some work and it's not very performant the way it is re-rendered every time the field is selected.

diegoliv:
Hey @Shelob9 I was thinking the same thing. I love Caldera Forms and it is my go-to form plugin for all projects. But I do think it would benefit from some UI / UX improvements. I was even thinking in open an issue here with some design suggestions to improve the UI. And also, I know that you guys started using Vue at some parts of the plugin. Have you ever though about adopting Vue for the form UI as well? I started working on the UI for the checkboxes yesterday and I must confess that I can clearly see the benefit of using Vue here.

Shelob9:
@diegoliv  Yes.

UI improvements are needed and we're planning on them now. As you've notice, we're mainly using Handlebars and jQuery, with newer stuff in Vue. A new UI for select fields should be a Vue component. Once this release is done, I'm going to redo all of our build tooling to use webpack (#2167) more intelligently. After that, this type of change -- replace one small part of the UI with a Vue component -- will be a lot simpler. My first project will be the magic tag auto-complete, then either calculations or conditionals. I'm setting that order based on looking at where the biggest memory losses were. If you want to propose a Vue component for a options, let's talk in #2327

classes/field/util.php
```diff
@@ -388,7 +388,7 @@ public static function get_default( $field, array  $form, $convert_opt = false )
 			$default = false;
 		}
 
-		if( $convert_opt && 0 === strpos( $default, 'opt' ) ){
+		if( is_array( $default ) || ( $convert_opt && 0 === strpos( $default, 'opt' ) ) ){
```
Shelob9: I am testing with a checkbox field with no default saved and the filter described here: https://github.com/CalderaWP/Caldera-Forms/pull/2322#issuecomment-368626113

I get this error
`Warning: strpos() expects parameter 1 to be string, array given in /app/wp-content/plugins/caldera-forms/classes/field/util.php on line 391`

BTW That makes 0% sense. I checked with xdebug that $default is an array, so you would think `strpos()` would not be executed, but it is. PHP is a troll sometimes.

Shelob9: There is only one form in my field, I watched every time it was executed. I'm going to stop arguing with the compiler and just fix this.

Shelob9:
@diegoliv #2356 is a new PR that includes everything you did and some additional work by me. I'm honestly not sure yet which one I want to merge.

Shelob9:
Github was smart enough to detect this got merged along with #2356 

Thanks @diegoliv !

diegoliv:
@Shelob9 glad to help! Happy to see this on the next release! 😄

Issue #2299, opened 2018-02-09
Breadcrumbs are hidden on form submit, not success.


Shelob9:
Submit form with validation error -- conditional fail processer is helpful for this. See attached form.

When fixing #1702 I did this:

https://github.com/CalderaWP/Caldera-Forms/commit/5c1e2eef1dadf47444c36e2c3d7c6b80a929a270#diff-376e984735c05c485ba5e75257c9c524R330

Breadcrumbs hidden after submission! Taco emojis, except that event cf.submit.event happens when it's submitted, not when it's successful.

**https://secure.helpscout.net/conversation/520071038/7512?folderId=832038**

Shelob9:
Pretty sure what we need to do here is switch to the cf.validate.FormSucces event.

Shelob9:
Test `2299 CF Feature: Multi-page create@x dev@1.5.9` is recorded and passed for this.

Shelob9:
Close via 72fd503 #2332

PR #2332, opened 2018-02-28
 Don't hide breadcrumbs until form submits correctly

Shelob9:
#2299


Shelob9:
There are extra commits in here, I may need to rebase or redo.

Issue #2337, opened 2018-03-02
WooCommerce messes up nonces for logged-out user


ictbeheer:
## What Version Of Caldera Forms, WordPress and PHP Are You Using?
WordPress Version: 4.8.5PHP Version: 7.0.27MySQL Version: 5.6.39Caldera Forms Version: 1.5.9.1WP_DEBUG:

## Does Your Issue Persist When You Disable All Other Plugins and Switch To The Default Theme?
No, because the issue is caused by WooCommerce

## What Is The Unexpected Behaviour?
The problem from #894 prevents logged-out users with items in their  WC cart from submitting Caldera Forms:

> When a user was logged out, but has items in their cart WooCommerce does a check and sets the user ID for non logged in user to be the session ID if one is set. And so the nonce validation doesn't work.

The fix from #894 is no longer working because the nonce action is no longer `called caldera_forms_front `but is now `caldera_forms_front_{$formid}`.

Shelob9:
PR #2338 should resolve this.

Issue #894, opened 2016-10-19
WooCommerce when user is not logged in, but items in carts messes with our nonce validation


Shelob9:
When a user was logged out, but has items in their cart WooCommerce does a check and sets the user ID for non logged in user to be the session ID if one is set. And so the nonce validation doesn't work.

This is a) real bad for us b) makes me wonder if I can have some of what they were smoking when they did this.

Workaround:

```
add_filter('nonce_user_logged_out', function($uid, $action) {
  if ($uid && $uid != 0 && $action && $action == 'caldera_forms_front') {
     $uid = 0;
  }

   return $uid;
}, 100, 2);

```

Shelob9:
Close via 	e574d15

PR #2338, opened 2018-03-02
Update how WooCommerce Nonce compatibility works

ictbeheer:
This is a regression fix for #894 which fixes the problem that WooCommerce causes by manipulating the nonce for logged-out users.

#2337


includes/functions.php
```diff
@@ -64,7 +64,7 @@ function caldera_forms_fix_license_manger_link(){
  */
 function caldera_forms_woo_nonce_fix( $user_id, $action) {
 	if ( class_exists( 'WooCommerce' ) ) {
-		if ( $user_id && $user_id != 0 && $action && $action == 'caldera_forms_front' ) {
+		if ( $user_id && $user_id !== 0 && $action && strpos( $action, 'caldera_forms_front' ) !== false ) {
```
Shelob9: @ictbeheer We use yoda conditionals. The old code is wrong in this manner.

`if ( $user_id &&  0 != $user_id && $action && 0 === strpos( $action, 'caldera_forms_front' )  )` would be more appropriate. Also, using `0 === strpos(`... ensures it is the start of the string `$action`

Shelob9:
@ictbeheer Thanks for your pull request. I left a comment on code style, but agree with your assessment of the bug/fix. If you can please update your PR, I will merge and include in 1.6 release.

ictbeheer:
@Shelob9 thanks for your review, I have updated the PR. I do have an additional question though. As Caldera Forms uses the WP Rest API should we also check for the `wp_rest` nonce?

Shelob9:
@ictbeheer Thanbks for the update, I will merge now.

>As Caldera Forms uses the WP Rest API should we also check for the wp_rest nonce?
I am very uncomfortable changing how that nonce works. I'll listen to a strong case for why I should, but the risk of making that nonce behave not as anyone would assume it does is scary to me.

Also, WooCommerce should play nice with that nonce, it's in WordPress core. Caldera Forms shouldn't have to work around WooCommerce like this, but whatever they are bigger and it's easier for us to work around them -- though it leads to having to do things like this.

Issue #2349, opened 2018-03-09
Integrate Freemius Insights


Shelob9:
From Vova: https://gist.github.com/vovafeldman/0f4c53175c5b46057b6772d1ee2677cb

Related: #307

Shelob9:
close via #2350 9b5cadb

PR #2350, opened 2018-03-09
Integrate Freemius insights

Shelob9:
#2349


caldera-core.php
```diff
@@ -75,7 +75,10 @@ function caldera_forms_load(){
 	include_once CFCORE_PATH . 'ui/blocks/init.php';
     include_once CFCORE_PATH . 'includes/cf-pro-client/cf-pro-init.php';
 
-	/**
+    //Initialize Freemius
+    add_action( 'caldera_forms_includes_complete', array( 'Caldera_Forms_Admin', 'get_freemius' ) );
```
vovafeldman: @Shelob9 it's important that Freemius will be executed right when the plugin's main file is being included, otherwise, the uninstallation event, for example, may not work. Many devs are initiating their logic using `'plugins_loaded'` or other filters/actions, I just want to make sure it's not the case here. P.S. My GitHub notifications aren't working so ping me when you reply.

Shelob9: @vovafeldman That is effectively what is happening here. This comment is in the main file, in a function hooked to `plugins_loaded`. The `caldera_forms_includes_complete` action is dispatched at the end of that function. 

If I put it any earlier, Caldera Forms' autoloader wouldn't be registered.

Shelob9:
@vovafeldman Can you please take a quick look at our activation hook and make sure that's not a problem? We do not use a deactivation hook.

https://github.com/CalderaWP/Caldera-Forms/blob/master/caldera-core.php#L39

vovafeldman:
@Shelob9 the activation is fine.

caldera-core.php
vovafeldman: @Shelob9  we use it to send the information captured on the feedback form. There's additional logic that needs to be executed before (don't remember what exactly). Anyhow, when I was testing Caldera Forms with Freemius to make sure everything will work fine, I appended the integration snippet at the end of the plugin's main file and everything worked fine: 
image

So you can either do the same or just put the snippet in a separate file and include it right away.

Shelob9: @vovafeldman Is there a functional reason that the function has to be called `cf_fs()` I can add a function in that file called `caldera_forms_freemius_init`.

vovafeldman: @Shelob9 this is just the shortcode that I used :) You can name it whatever you want, just keep a unique naming.

Issue #2195, opened 2018-01-03
Structure of API responses for entry routes is sub-optimal 


Shelob9:
Same issue as #2194

Issue #2194, opened 2018-01-03
Structure of API responses for form routes is sub-optimal


Shelob9:
Right now JSON for /forms parses to something like this: `"cf124" : { "ID" => "CF124} }` we want `{ [ { "ID" => "CF124} ]` for single we want `{ "ID" => "CF124}`

Shelob9:
Need to fix this before doing #857 which is now priority.

PR #2457, opened 2018-04-23
Use index_by_id param in forms and entries responses to return array if needed

Shelob9:
For backwards compat reasons, default behavior is not changed, new param is added

#2194 - For an ARRAY: `https://<url>/wp-json/cf-api/v2/forms?index_by_id=1`

#2195 - For an ARRAY `https:/<url>/wp-json/cf-api/v2/entries/<form-id>?index_by_id=1`


Shelob9:
I'm closing this. Do not want to change API v2, that's what api v3.

Shelob9:
I want #2194 not #2195 I will look at this today.

Issue #2429, opened 2018-04-13
Provide data to GDPR export/delete/annoymize requests from core


Shelob9:
Filters: 
* `wp_privacy_personal_data_exporters`

https://danieliser.com/practical-guide-to-gdpr-compliance-for-wordpress-plugin-and-theme-developers/

- [x] Data Exporter: https://gist.github.com/danieliser/5ae69ed33491a43ddd62fe05efb6d4c0#file-plugin-privacy-user-data-exporter-php
- [x] Eraser : https://gist.github.com/danieliser/b6d2a57b795349af4e5127d9081dc1c2#file-plugin-privacy-user-data-eraser-php

nicomollet:
Very much appreciated. Gravity Forms, Ninja Forms and Formidable Forms don't seem to care about this tool.

Shelob9:
Close via 656bdd8

PR #2507, opened 2018-05-16
Add privacy/data settings screen and integrate with WordPress core private data exporters and deleters

Shelob9:
#2429


Shelob9:
For future readers/ future me: the solution to my concern about not loosing saved data was to create caldera_forms_editor_extra_fields filter (21c139a) https://calderaforms.com/doc/caldera_forms_editor_extra_fields/

Issue #2524, opened 2018-05-20
Don't ask for slug or name for certain field types


iamntz:
Some fields asks for an slug and name even they aren't useful in that context: separator and HTML content being the culprits.

I understand that in some cases the slug can be useful (e.g. to show/hide based on a condition), but i guess these values can be pre-populated with an `uniqid()` or something similar.

Shelob9:
Close via 6b6ad4a  Thanks @iamntz

PR #2525, opened 2018-05-20
For certain fields type auto-populate field slug & name

iamntz:
Quick fix for #2524

### DIFF
--- a/assets/js/edit.js
+++ b/assets/js/edit.js
@@ -199,7 +199,6 @@ jQuery(document).ready(function($){
 
         // remove not supported stuff
         if(fieldtype_defaults[select.val() + '_nosupport']){
-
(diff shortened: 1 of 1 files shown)


Shelob9:
Thanks for the pull request @iamntz This looks good. I need to test before merging, but I think it is a sensible change.

assets/js/edit.js
```diff
@@ -261,6 +260,13 @@ jQuery(document).ready(function($){
                 }
             });
         }
+
+        if (['html', 'section_break'].indexOf(select.val()) >= 0) {
+            var label = parent.find('.field-label');
```
Shelob9: @iamntz Small point, but I prefer that a variable that holds a `jQuery` instance be prefixed with `$`. Not a blocker, edit.js does not follow this convention well.

iamntz:
I updated the var declaration

Issue #2458, opened 2018-04-25
Suddenly unable to see select boxes - by user role


ferbit:
Transferring this issue to Caldera Forms from pods, per Jimtrue.  Hope you guys can help! This started happening right after updating the Caldera Forms plugin this morning.

https://github.com/pods-framework/pods/issues/4863

PR #2535, opened 2018-05-22
Removed redundant old code from build/js/shortcode-insert.min.js .

### LINKED ISSUE #2448: Uncaught TypeError: Cannot read property 'data' of undefined
Short Debug Information
WordPress Version: 4.9.5PHP Version: 5.6.35MySQL Version: 5.6.36Caldera Forms Version: 1.6.0WP_DEBUG:

There seems to be a conflict with the Pods plugin (I have version 2.7.1). When I deactivate Pods, the javascript error disappears. I also have Caldera Easy Pods installed (version 1.1.8), Caldera Forms version 1.6.0.

I filled out the form and clicked Submit. Nothing happened.

No PHP errors or notices related to this.

The error below is from the Chrome console. The error comes up when entering a text in one of the form fields. 

Uncaught TypeError: Cannot read property 'data' of undefined
    at Object.error (ajax-core.min.js?ver=1.6.0:1)
    at Object.request_error (jquery-baldrick.min.js?ver=1.6.0:1)
    at c (jquery-baldrick.min.js?ver=1.6.0:1)
    at Object.error (jquery-baldrick.min.js?ver=1.6.0:1)
    at i (jquery.js:2)
    at Object.fireWith [as rejectWith] (jquery.js:2)
    at y (jquery.js:4)
    at Object.abort (jquery.js:4)
    at Object.request (jquery-baldrick.min.js?ver=1.6.0:1)
    at c (jquery-baldrick.min.js?ver=1.6.0:1)

0xLBF:
Two versions of shortcode-insert.js were compiled into assets/build/js/shortcode-insert.min.js (its size became 4.82KB instead of about 2.41KB), removed the second one.

The second one was an older version that did not include the #caldera-forms-form-insert check for wp.mce.views, and it appeared to be a re-minified version of assets/js/shortcode-insert.min.js, deleted that as well. Looks like something built both assets/js/shortcode-insert.js and assets/js/shortcode-insert.min.js into one final assets/build/js/shortcode-insert.min.js.

This fixes the bugs that reoccur without the #caldera-forms-form-insert check. #2458 has the exact same JavaScript errors I had but in my case they happened in the Pods Template editor.

TODO: Other min.js files may have a similar problem.

The final solution should probably involve rebuilding the .min.js files properly rather than editing them like I did.


Shelob9:
Good catch. I opened #2537 for this issue.

PR #2548, opened 2018-05-25
Extracted fields to each own method

iamntz:
Although this seems a HUGE commit, is actually a small one: I only extracted each field config to it's own method, for a better readability.

### DIFF
--- a/classes/fields.php
+++ b/classes/fields.php
@@ -99,543 +99,699 @@ class Caldera_Forms_Fields {
 	 * @return array
 	 */
 	public static function internal_types() {
+		$suffix = defined('SCRIPT_DEBUG') && SCRIPT_DEBUG ? '.min' : '';
+
 		$internal_fields = array(
 			//basic
-			'text'             => array(
-				"field"       => __( 'Single Line Text', 'caldera-forms' ),
-				"description" => __( 'Single Line Text', 'caldera-forms' ),
-				"file"        => CFCORE_PATH . "fields/generic-input.php",
-				"category"    => __( 'Basic', 'caldera-forms' ),
-				"setup"       => array(
-					"template" => CFCORE_PATH . "fields/text/config.php",
-					"preview"  => CFCORE_PATH . "fields/text/preview.php"
-				),
-
-			),
-			'hidden'           => array(
-				"field"       => __( 'Hidden', 'caldera-forms' ),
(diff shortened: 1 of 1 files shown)


Shelob9:
@iamntz I love this idea. I think we need to think it through before making the change. I opened #2549 to discuss.

classes/fields.php
```diff
@@ -99,543 +99,699 @@ public static function not_support( $type, $not_support ){
...
+			'states'             => self::get_states_field(),
+
+			//discontinued
+			'recaptcha'          => self::get_recaptcha_field(),
+		);
+
+		return $internal_fields;
+	}
+
+	public static function get_text_field()
```
Shelob9: All methods need inline docs: https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#1-functions-class-methods

classes/fields.php
```diff
@@ -99,543 +99,699 @@ public static function not_support( $type, $not_support ){
...
-					"not_supported" => array(
-						'hide_label',
-						'caption',
-						'required',
-						'entry_list'
-					)
+		);
+	}
+
+	public static function get_hidden_field()
```
Shelob9: I think that the names of methods should use the word config as the code base/ documentation/ Josh consistently uses the term "field config". The class name says fields, so `hidden_config` or `get_hidden_config()`

classes/fields.php
```diff
@@ -99,543 +99,699 @@ public static function not_support( $type, $not_support ){
...
+	{
+		return array(
```
Shelob9: arrays should use bracket syntax, but don't worry #2362 will fix that for us. I know this is copying legacy code, don't worry about it.

classes/fields.php
iamntz: I think `get_` should remain (because it... uhm... _gets_ the config). I'm changing everything to `get_*_config()`

classes/fields.php
iamntz: While this may be useful in other context, I think it will be just noise in the code. methods are about 10 lines long and kind of self-documented...

iamntz: I think the only _potential_ useful would be the `@version` comment, but even that i think that adds only very little value.

iamntz: I've added the docblocks too.

PR #2604, opened 2018-06-09
Added a wrapper to the form pages

iamntz:
I found myself in need to have form pages wrapped in a element, with nothing else (think: carousel pages)

So this will add the ability to wrap all pages in a div and leave everything else outside (e.g. pagination, hidden fields and such)

### DIFF
--- a/classes/core.php
+++ b/classes/core.php
@@ -4698,7 +4698,10 @@ class Caldera_Forms
 			}
 
 
-			$out .= $form['grid_object']->renderLayout();
+			out .= sprintf('<div class="%s">%s</div>',
+				apply_filters('caldera_forms_grid_wrapper_class', 'cf-grid-wrapper', $form),
+				$form['grid_object']->renderLayout()
+			);
 
 			$out .= "</" . $form_element . ">\r\n";
 		}


classes/core.php
```diff
@@ -4672,7 +4672,7 @@ static public function render_form($atts, $entry_id = null, $shortcode = null)
 				// retain query string
 				$qurystr = array();
 				parse_str($_SERVER['QUERY_STRING'], $qurystr);
-				$out .= "<span class=\"caldera-grid\"><ol class=\"breadcrumb\" data-form=\"caldera_form_" . $current_form_count . "\" id=\"caldera-forms-breadcrumb_" . $current_form_count . "\">\r\n";
+				$out .= "<div class=\"caldera-grid\"><ol class=\"breadcrumb\" data-form=\"caldera_form_" . $current_form_count . "\" id=\"caldera-forms-breadcrumb_" . $current_form_count . "\">\r\n";
```
Shelob9: @iamntz In isolation, you are correct. This is more semantic.

I am fairly opposed to this change because a change in element type can effect CSS targeting and JavaScript bindings. How series of a concern is that? How many of the 100-200k sites running CF does that matter to? No idea.

But I could be convinced that I'm over-reacting.

Shelob9:
@iamntz I'm not sure I agree here with the need and I'm concerned, though those concerns may be unwarranted. Can you please open an issue explaining why this change is needed?

Shelob9:
Also, this is going to require developing additional acceptance tests for multi-page forms to make sure all the different validation steps are completed properly.

PR #2605, opened 2018-06-09
Apply wpautop to front-end display of HTML fields.

iamntz:


Shelob9:
Good catch. I opened #2619 for this and edited the title to reflect the contents of the change.

fields/html/field.php
```diff
@@ -21,7 +21,7 @@
 
 }else{
 	$html_template = $field[ 'config' ][ 'default' ];
-	$content = do_shortcode( Caldera_Forms::do_magic_tags(  $html_template   ) );
+	$content = do_shortcode( Caldera_Forms::do_magic_tags( wpautop( $html_template ) ) );
```
Shelob9: @iamntz Not a blocker, but it concerns me as your  commit messages become part of our git history, that your commit message doesn't describe the reason for the commit, which happens here.

A better commit message would be something like
Applies wpautop formatting to html field front-end display #2619

Shelob9:
I am merging as failed job https://travis-ci.org/CalderaWP/Caldera-Forms/jobs/390118886 is a Travis issue. Other tests passed.

lyatziv:
This fix breaks any javascript that may be in a `$html_template`

Issue #2596, opened 2018-06-07
Add JavaScript unit testing with Jest


Shelob9:

Shelob9:
Close via 
69e7cfa

Issue #2564, opened 2018-05-31
Use @wordpress/data for state in privacy screen


Shelob9:
Currently using Redux in privacy settings screen, instead of using `wp.data` which is what is used for blocks. It will be simpler to use one thing. https://www.npmjs.com/package/@wordpress/data

Shelob9:
The https://www.npmjs.com/package/@caldera-labs/state is the first step to integrating.

Shelob9:
Close via 9164706

PR #2611, opened 2018-06-11
Use state and api libraries for block and privacy state and api

Shelob9:
#2564 

🌋 https://www.npmjs.com/package/@caldera-labs/state
🌋 https://www.npmjs.com/package/@caldera-labs/api-client

This pull request takes all the most recent JavaScript code I wrote, deletes it and then replaces it with code from two new libraries that are copy pasted from that code I just deleted. These new packages are:

* Have unit tests
* Are well documented. [For example](http://calderalabs.org/caldera-api-client/manual/overview.html#mocking-requests)
* Can be re-used without further cut and paste.
* Have no dependency on WordPress or Caldera Forms.
* [Helped me make a boilerplate to speed up creating these types of library](https://github.com/CalderaWP/caldera-npm-module-boilerplate/)

Goals:

- [x]  Prove the new libraries/ this approach to development.
- [x] Wrap all state management and API requests for blocks in new libraries
- [x] Wrap all state management and API requests for privacy screen in new libraries
- [x] Remove usage of `wp` global so code more is testable, more explicit.
- [x] Make the privacy screen tested to serve as reference for admin screen. Doable with #2596

So this is a lot of work to make no changes to functionality at all. But this is the new way, may it slowly consume all of the Javascripts in its well-tested and well-documented lava flow of glory. 🌋


Shelob9:
screen shot 2018-06-12 at 5 14 49 pm

Issue #2684, opened 2018-07-24
Support calculation value using caldera_forms_field_option_presets filter


mecachisenros:
## Do You Need Immediate Help?
No

## Is This A Feature Request?
Yes

## Description
Currently the `caldera_forms_field_option_presets` filter allows to add preset options like `label` or `value|label`, defaulting the `value` to the `label` if this is ignored.

This feature request is to support also the `calc_value` parameter in the format of `value|label|calc_value`, this ensures non breakage for plugins/implementation using the filter.

### Before
![before](https://user-images.githubusercontent.com/3038096/43169913-c6332eaa-8f9a-11e8-9fbb-a94d1bae0555.gif)

### After - Label only
![after_label](https://user-images.githubusercontent.com/3038096/43169934-db2cb808-8f9a-11e8-860b-e43e4539186f.gif)

### After - Value and label
![after_value_label](https://user-images.githubusercontent.com/3038096/43169968-fa2f2cf4-8f9a-11e8-94fc-00e451127cae.gif)

### After - Value, label and calculation value
![after_value_label_calc_value](https://user-images.githubusercontent.com/3038096/43169992-0c0fab56-8f9b-11e8-8960-50c27681e59d.gif)

Shelob9:
Close via #2685

Shelob9:
Thanks!

PR #2685, opened 2018-07-24
support calc_value using caldera_forms_field_option_presets filter

mecachisenros:
As per my description in #2684


Shelob9:
Thanks @mecachisenros this will be included in 1.7.3

mecachisenros:
Many thanks @Shelob9

PR #2691, opened 2018-08-07
Fixed a significant performance issue with select2 fields.

0xLBF:
#2697

The select2 fields were initialized over and over again whenever any other conditional field became visible. For large forms with many select2 and conditional fields, this could actually double the whole page load time.

This patch makes sure the select2 fields are initialized only once on load, and whenever they become visible if they're a conditional field.

TODO: The run_on_init_add function should probably go somewhere else so that it can be reused by custom fields and other user code ? Also, there might be even better ways of implementing this, for example we could avoid adding any event listeners for unconditional fields, but this already seems to be fast enough for now.

### DIFF
--- a/assets/js/state/state.js
+++ b/assets/js/state/state.js
@@ -250,6 +250,9 @@ function CFState(formId, $ ){
 			});
 			calcVals[id] = findCalcVal( $( document.getElementById( id ) ) );
 			self.mutateState([$field.attr('id')],$field.val());
+			$field.trigger('cf.bind', {
+				field: $field.attr('id')
+			});
 
 			return true;
 		} else {
@@ -333,6 +336,10 @@ function CFState(formId, $ ){
 
 
 					self.mutateState(id,val);
+					
+					$field.trigger('cf.bind', {
+						field: $field.attr('id')
+					});
 
 				});
 				return true;
--- a/fields/select2/field/field.php
+++ b/fields/select2/field/field.php
@@ -131,34 +131,8 @@ jQuery( function($){
 	var opts = {};
 	<?php } ?>
 
-	// run a function no more than once every time the field needs to be initialized
-	// i.e either on page load or when the conditional field becomes visible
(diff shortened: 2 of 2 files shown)


Shelob9:
Good find. 

>TODO: The run_on_init_add function should probably go somewhere else so that it can be reused by custom fields and other user code ? Also, there might be even better ways of implementing this, for example we could avoid adding any event listeners for unconditional fields, but this already seems to be fast enough for now.

The initialization function really should be inside of this closure: https://github.com/CalderaWP/Caldera-Forms/blob/master/assets/js/field-config.js

That way the shared logic can be in the scope of the closure and shared with other field init methods. If you look in that file, there is a method for most field types. They are all sharing the same state object and event emitter. Getting everything managed by those objects  -- state, events and field configs and eventually conditionals -- should allow for exactly these types of optimizations

0xLBF:
> That way the shared logic can be in the scope of the closure and shared with other field init methods

That's great for the internal / regular field types, but if I have a custom field type that's defined a separate WP plugin, that would need some changes to allow plugging additional custom initializers into that closure, reusing some methods of that closure, without changing the Caldera plugin code ?

Shelob9:
>the closure would need some changes to allow plugging additional custom initializers into that closure
Correct. 

What would ideally happen is that the custom plugin would subscribe to changes in the field.

* Event subscriber: https://github.com/CalderaWP/Caldera-Forms/blob/master/assets/js/state/events.js#L22
* The specific event subscriber/emitter instance for a specific form in front-end is contained in that form's state object: https://github.com/CalderaWP/Caldera-Forms/blob/master/assets/js/state/state.js#L191
* That state object is acessible via `window.cfState[ form_id ]` where form_id is the `<form>` id attribute, not form ID.

@0xLBF Feel free to DM me in WordPress slack - I am shelob9 there. Happy to go over details.

Shelob9:
Thanks @0xLBF !

Issue #2652, opened 2018-06-29
Losing validation of Phone Better field in conditional logic 


salihkulangara:
When we use the Phone Number(Better) field in a conditional logic, first time when the field show up it will have all the validation and country flags and all, but if we switch between the conditions, lose the validations and county flags, to better explain please see the attached gif.

![caldera-issue](https://user-images.githubusercontent.com/6852780/42073102-1d314764-7b82-11e8-9cb4-377d75bcde81.gif)

The Phone Number(Basic) also had the issue - losing the field masking and validations, still version Version 1.7.1.4, but in Version 1.7.2, it seems fixed, For Phone Number(Better) field issue still exists.

Wordpress:  4.9.6
Caldera Forms: 1.7.2
PHP: 5.6.31

Shelob9:
@salihkulangara Thanks for reporting this. 

Notes:
- This likely has the same solution as  #2612
- The handler function for phone_better fields is here - https://github.com/CalderaWP/Caldera-Forms/blob/master/assets/js/field-config.js#L383
- That file is contacted into caldera-forms-front.min.js

Shelob9:
New test 2652 CF Field: Phone Better created@1.7.3 proves this is a real issue.

The init function *is* running after the field is unhidden, but it still doesn't reload the functionality of this field type. Confused now, I will look further.

mecachisenros:
@Shelob9 this should be fixed via #2711, I've added a brief explanation in the PR.

### Demo
![phone_init_validation](https://user-images.githubusercontent.com/3038096/44633855-73e78800-a988-11e8-98dc-17cd9c63358c.gif)

Shelob9:
Fixed via #2711

salihkulangara:
Great

Issue #2703, opened 2018-08-19
Phone (better) field displaying 'Not a number' error message


mecachisenros:
## Do You Need Immediate Help?
No

## Is This A Feature Request?
No

## What Version Of Caldera Forms, WordPress and PHP Are You Using?
WordPress Version: 4.9.8
PHP Version: 7.0.31
MySQL Version: 5.5.5
Caldera Forms Version: 1.7.2

## Does Your Issue Persist When You Disable All Other Plugins and Switch To The Default Theme?
Yes

## What Is The Unexpected Behaviour?
The Phone (better) field displays a `Not a number` error message when the field is empty, with both _Use Country Code_ setting enabled or disabled.

If you click on the field or navigate with tab key and leave the field empty you get a `Not a number` validation message, the validation is performed on `change`, `blur/focus`, and `submit` events, so even if don't _touch_ the field, the error displays briefly on form submission as well.

I would expect no validation error if the field is empty. 

## What PHP Errors Have You Logged While Reproducing This Bug?
No PHP errors.

## What JavaScript Errors Have You Seen While Reproducing This Bug?
No JS errors.

I propose to change the [validation function](https://github.com/CalderaWP/Caldera-Forms/blob/master/assets/js/caldera-forms-front.js#L6054) to handle both cases, empty field, and empty field with country code (_Use Country Code_ enabled).
```
 var validation = function () {
     reset();
     var valid;
     var value = $.trim($field.val()); // field value
     if (value) {
         if ($field.intlTelInput("isValidNumber")) {
             valid = true;
         } else {
             valid = false;
         }
     }

     var message;
     var errorCode = $field.intlTelInput("getValidationError");
     var selectedCountryData = $field.intlTelInput("getSelectedCountryData"); // selected country data

     if (0 == errorCode) {
         valid = true;
         message = '';
     } else if ( value ==  "+" + selectedCountryData.dialCode ) { // Use Country Code enabled (this settings fills in the dial code as value)
         	valid = true;
         	message = '';
     } else if ( ! value ) { // empty value
         	valid = true;
         	message = '';
     } else {
         if ('undefined' != field.messages[errorCode]) {
             message = field.messages[errorCode]
         } else {
             message = field.messages.generic;
         }
     }


     handleValidationMarkup(valid, $field, message, 'help-block-phone_better');
     return valid;
 };
```

### Before
**Use Country Code setting disabled, initial state**
[...]

PR #2704, opened 2018-08-19
fix Phone (better) field 'Not a number' validation error message

mecachisenros:
Issue described here #2703


Shelob9:
@mecachisenros Thanks for this PR. The file you're modifying is a generated file, so if I merged this, it would be immediately lost. That function is here:
https://github.com/CalderaWP/Caldera-Forms/blob/master/assets/js/field-config.js#L383

You can get CF to load that file with this filter: https://calderaforms.com/doc/caldera_forms_render_assets_minify/

Shelob9:
Hopefully this can fix #2652 as well

mecachisenros:
@Shelob9 thank you for the heads up, I did not see the `/*! GENERATED SOURCE FILE */` 😄 
I've amended the commit, so this should be ready to merge.

Unfortunately it does not fix #2652, I'll have a look see if I find a fix for it.

Shelob9:
@New0 Can you please test this now that #2711 is merged ?

Shelob9:
@mecachisenros @New0 It looks like the conditional logic is not reinitalizing after the field is hidden, then unhidden by conditional logic.

Two theories:
1) The init function is not running after the field is re-added, which should happen beacuse of this binding to `cf.add` event:
https://github.com/CalderaWP/Caldera-Forms/blob/5eb23f08e14e453702f063630d2870eff10b5df9/assets/js/field-config.js#L450

2) Something changed re: WHEN the `cf.add` event runs, so it doesn't reinit the field, which I don't think so but FML if that's the problem.

3) It's running the init() function with bad arguments or before the field is back on the DOM after conditional logic unhides the field.

Shelob9:
@New0  I pushed commits for the issues we found while discussing this - validation not triggered after unsave and button not releasing.

Can you please re-review?

New0:
@Shelob9 I ran tests with and without assets minified and the field had the same behavior we saw on your screen.

- The 'not a number message shows correctly before the condition hides the phone field, but when the condition shows back the phone field ->
    1 the 'not a number' is not displayed
    2 the submit button was re-enabled
    3 I was able to submit the form

- The only difference I found when setting the phone field as required is that the submit button is disabled before the phone field gets hidden, when the phone fields is shown back, the submit button was enabled again and I was able to submit.

Shelob9:
@New0 Check this out:
https://github.com/CalderaWP/Caldera-Forms/blob/master/assets/js/conditionals.js#L308-L315

`cf.add` gets triggered and THEN the value gets restored to the field. So sequence of events is now:

1) Place field on DOM with NO value set
2) Trigger `cf.add`
3) Rerun validation
4) Reset the stored value to the field.

If 3 and 4 were reversed this would make sense, as is its madeness. I'm afraid to modify conditionals.js

Shelob9:
My solution is when cf.add is triggered, to bind to the state update event, trigger init and then remove that event. That way the initialization and validation runs once the value is reset. I remove it to prevent it from reinitializing on every change event to the phone field.

Shelob9:
@New0 Please review again. Also, do you have a good solution for the merge conflicts in generated files?

New0:
@Shelob9 I added the validation on the submit button click, I think the validation we intend to do at field-config.js l 441 happens too late, the form submit event already started

This works for submission but that can create issue for multi page forms as it doesn't work for next page button click event.

Shelob9:
@New0 Can you solve this by adding events on page forward and back? if so, please add those events.

New0:
@Shelob9 I couldn't find a solution by triggering a custom event during the .on click event that is already set up. 
I set a .click event in the field-configs, I tested with multiple phone fields and multiple forms in the page. Could you review ?

kcristiano:
@Shelob9 This is quite a critical fix for a number of our sites, can you let us know what (if anything) we can do to help move this along?

New0:
@kcristiano testing this branch on a staging environment of your websites and reporting back would help. To do so, it would require to rebuild the dependencies using composer.

Otherwise, wait Josh review and test the develop branch when this PR will be merge.

New0:
@kcristiano this is a development branch, it is not ready for production. The working patch was introduced yesterday, I strongly advise to switch back to master branch for production and wait for this branch to be merged to master, this is when we'll run all final tests and push the master branch to WordPress.org repository

kcristiano:
Thanks @Shelob9 I was not looking to give anyone grief on the time it took, I know it was complicated and appreciate the work.  I was trying to explain why we needed to move the PR to production.  I much prefer to run on the stable branch.

PR #2711, opened 2018-08-26
fix phone (better) field init/validation DOM element reference

mecachisenros:
Fix for #2652

### Explanation
Using field conditionals the _conditional_ elements are destroyed/removed/created based on the condition, the `phone_better` field holds a `$field` reference to the DOM element when the form is created. 

The `init` and `validation` functions are triggered more than once when conditionals are present, the reference must be updated in both functions as the `$field` variable holds a reference to a non existent element once the element is destroyed/removed (ie condition changes).


Shelob9:
@New0 I made test 2652 CF Field: Phone Better created@1.7.3 for this, can you please pull this down, re-compile and test?

Shelob9:
Thanks @mecachisenros  and @New0

PR #2721, opened 2018-09-09
Fix typo in previews. <lable> should be <label>

shawnhooper:
The preview.php files for each control have a typo in the <label> HTML element.  

Also, Hi Josh.


Shelob9:
Hi Shawn. 

Two issues that concern me here:

0) Something makes me think this was an intentional typo, but not sure.
1) I found 'lable' used 6 times in assets/js/edit.js Can you change those as well please?
2) You submitted against master branch. I fixed.

New0:
Thank you again for the PR @shawnhooper 
I think it would be worth fixing the label tags, I tested local tests and that passed ( after merging current develop branch into shawnhooper:fix/labels ), @shawnhooper could you update to current develop?
The main issue I see is if custom styles or js targeted lable instead of label, @Shelob9 what do you think?

PR #2726, opened 2018-09-13
Compare saved user ID correctly 

### LINKED ISSUE #2725: Non admin-users can not edit their own entries in front-end.
When editing a saved entry, if user is not an admin, then it will fail here:

https://github.com/CalderaWP/Caldera-Forms/blob/master/classes/core.php#L4329

Even if `$details[ 'user_id' ]` is `==` to `current_user_id()` beacuse `$details[ 'user_id' ]` is a string. Casting or strict equality check is required here.

Shelob9:
Fixes #2725


Shelob9:
The customer whose support ticket was from reported the issue was fixed.

I also confirmed the fix locally.

Shelob9:
@New0 Please review again and merge

Issue #2645, opened 2018-06-27
cannot save the "Enable Enhanced Delivery" setting. I click it to deselect it, but when I come back to it, it is reenabled.


cicichirinos:
see https://secure.helpscout.net/conversation/606644746/10702/?folderId=832038

Shelob9:
Should be fixed via #2645

PR #2738, opened 2018-09-27
#2645 Pro settings not saving in Form's tab view

New0:
Caldera Forms Pro Settings were not saving in the Form pro tab view: 
- This because the Form object was set by a select option field in the Caldera Forms Pro page 
- Passed the "formScreen" util in the STATE list in "clients/pro/store/index.js" in order to get current form ID when form object was not defined by the select field
- Used a new  conditionally defined "singleForm" object for form in "clients/pro/components/FormSettings/Form.vue"

### DIFF
--- a/clients/pro/components/FormSettings/Form.vue
+++ b/clients/pro/components/FormSettings/Form.vue
@@ -1,7 +1,7 @@
 <template>
 	<div>
 		<div class="caldera-config-group">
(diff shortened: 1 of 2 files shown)


Shelob9:
@New0 This looks good. I will test today.

Shelob9:
@New0 This is ready to merge.

New0:
This fixed #2645

Issue #2735, opened 2018-09-23
Consider add a comment about PII string to help translators


Rahmon:
It's not easy to translators know what "PII" meaning so I think it's a good idea add a comment. 

https://github.com/CalderaWP/Caldera-Forms/blob/2cdb95b7f258b6567f0b570da526ab4bd6d58ec3/classes/admin/privacy.php#L78

Shelob9:
Fixed via #2746  c175c2c

Thanks for the patch.

PR #2746, opened 2018-10-01
Add a comment about PII string to help translators

Rahmon:
It's not easy to translators know what "PII" meaning so I think it's a good idea add a comment.

Closes: #2735


Shelob9:
Thanks @Rahmon !

Issue #2658, opened 2018-07-05
Conditions do not save


DeanBu:
## What Version Of Caldera Forms, WordPress and PHP Are You Using?
WordPress Version: 4.9.6
PHP Version: 7.1.7
MySQL Version: 5.6.37
Caldera Forms Version: 1.7.2

## Does Your Issue Persist When You Disable All Other Plugins and Switch To The Default Theme?

Yes.

## What Is The Unexpected Behaviour?

A form was created which includes Field Type: 'Toggle Field' to set up different payment options for users: Credit card, Check or wire transfer.

When the user selects any payment method that is not credit card, Stripe should not be available to use. 

This was set up using a condition set in the Stripe Processor, but the condition does not retain the information after saving/when the form is refreshed.

The redirect processor conditions has also been tested and this also does not retain conditions after refreshing.

Shelob9:
I tested. These are my findings:

* A form with just two toggle switch fields and a redirect that has a toggle switch-based conditionals does save/refresh properly.
* I imported the customers form into a clean site without CF Stripe, I setup a redirect produced with toggle switch based conditionals. When I saved, they were gone, customer issue reproduced. I can't get any type of conditionals to save in the customer form with CF Stripe active.

Clearly I was full of shit when I told @DeanBu this was clearly a CF, not CF Stripe bug or when I told @shelob9 this would be an easy fix.

This is not the only Helpscout ticket having this issue with our payment add-ons, unlear if this is in v1 scope or not.

Shelob9:
This is a duplicate of #2512 

The workaround is here: https://github.com/CalderaWP/Caldera-Forms/issues/2512#issuecomment-405936147

PR #2753, opened 2018-10-05
Remove call to rebind_field_bindings on change of  .set-system-variable

### LINKED ISSUE #2512: Conditionals From Auto-Responder Do Not Save If There Are Variables In Use
Version 1.6.2 (but WP says 1.6.3)

YES, but disapears when I rollback to previous version

After Update, the Conditions in Auto Responder are Blank!

After rollback to v 1.6.1.1 everything is back to normal. But you've lost most of the settings if you saved the form in the meantime. If you haven't saved the template in v 1.6.2, the settings seems to be back completely.

Not sure, wheather the condition actualy worked on fontend (until saving the form, it might be just UI problem). Didn't have the nerve to test it more, it happened to me on live site.

BTW: not sure, what version it actually is: the changelog says 1.6.2, WP says 1.6.3.

No PHP error as far as I remember.

Unfortuately I rolled back to previous version, can't tell now.

New0:
This Branch is meant to test properly the effect of Removing call to rebind_field_bindings on change of  .set-system-variable
It fixes #2512 and maybe #2658

I understand this function is supposed to bind the option like system vars, magic tags and variables in the options lists in different fields of the Form's panel.

In the precise case of a variable, it will be immediately available in options during creation and should to during edition, but it is not, when the variable is edited, it requires to save and reload the page for the variable to be correct in fields options.

I suggest we test this as a PR to fix #2512 and open an issue for variable edited to be immediately available in options


New0:
The changes to variables actually appear immediately in the options, my bad ( my first test was wrong ).

Calling rebind_field_bindings on change here is actually working so I need to prevent this change action to be triggered on page load.

New0:
I removed the function that triggered the change event, @Shelob9 let me know if I'm wrong, I don't think it's necessary to trigger the change event like it was.

I also realized that variables broke the Gravatar option field in Form Settings, this PR fix this issues to.

Issue #2757, opened 2018-10-08
Travis CI tests failing on dependency missing


New0:
Example is PR #2755
Rebuild dependencies to run tests again.

PR #2758, opened 2018-10-08
Rebuild dependencies

New0:
Travis tests failing on missing dependencies. This is work on #2757


Shelob9:
@New0 TL;DR `composer require calderawp/caldera-containers` and I like your fix for phpunit.

> the new vendor/bin/phpunit file added a check to see if we are running in Cygwin by checking for cygpath program. That breaks the path to ../phpunit/phpunit/phpunit which exists but fails to be called
🤷‍♀️ This makes total sense to me:

> If I overpass the first issue by a direct call to ../phpunit/phpunit/phpunit 
Fuck it, do that. Let's get this release out.

> I run into an issue opening 
> calderawp/caldera-forms-query/src/CalderaFormsQueries.php the file exists but I found that it is using "use calderawp\CalderaContainers\Service\Container;" and that path doesn't exist, I think it is 

This *should* be fixed by composer updating Caldera Forms. The package `calderawp/caldera-containers` is a dependency of `calderawp/caldera-forms-query` See: https://github.com/CalderaWP/caldera-forms-query/blob/master/composer.json#L20

I actually like the idea of making `calderawp/caldera-containers` a requirement of Caldera Forms in its composer.json because more packages are going to be brought in to Caldera Forms that use `calderawp/caldera-containers` for their service container so we should want Caldera Forms to set the minimum version of that dependency.

New0:
@Shelob9 I think I wasn't precise enough, the calderawp/caldera-container exists but in caldera-forms-query, calderawp/caldera-forms-query/src/CalderaFormsQueries.php is calling "use calderawp\CalderaContainers\Service\Container;" but as the container.php file exists inside src and src/service folders, I think the src is missing from the path.

That made me think this might be a typo in calderawp/caldera-forms-query/src/CalderaFormsQueries.php l5 where we should rebuild to call "use calderawp\CalderaContainers\src\Service\Container;" or "use calderawp\CalderaContainers\src\Container;"

Shelob9:
@New0 `use calderawp\CalderaContainers\Service\Container;` is correct. `src` is the root of the PSR-4 namespace. See:

* https://github.com/CalderaWP/caldera-containers/blob/master/src/Service/Container.php#L4
* https://github.com/CalderaWP/caldera-containers/blob/master/composer.json#L19

Shelob9:
4882eb9  - Add a test to check if query builder and container dependencies work.

589f878 - Require caldera-containers. 

Let's see if Travis likes it.

Shelob9:
OK, check this out. It's not that the file doesn't exist. It doesn't exist in the location Composer thinks it should be in. It exists where we think it should go.

`PHP Fatal error:  require(): Failed opening required '/home/travis/build/CalderaWP/Caldera-Forms/vendor/composer/../calderawp/caldera-forms-query/src/CalderaFormsQueries.php' (include_path='.:/home/travis/.phpenv/versions/7.0.25/share/pear') in /home/travis/build/CalderaWP/Caldera-Forms/vendor/composer/autoload_real.php on line 66`


It's trying to include `/home/travis/build/CalderaWP/Caldera-Forms/vendor/composer/../calderawp/caldera-forms-query/src/CalderaFormsQueries.php` It should be trying to load `/home/travis/build/CalderaWP/Caldera-Forms/vendor/calderawp/caldera-forms-query/src/CalderaFormsQueries.php`

Notice the `/vendor**/composer/../calderawp/**. That's because it thinks its in the autoloader's namespace or some shit. I promise I know how this thing works.

New theories:
1) https://github.com/CalderaWP/caldera-forms-query/issues/6
2) We're missing a `\` somewhere before a namespace.

Shelob9:
@New0 `src` is empty:

screen shot 2018-10-05 at 1 30 38 pm

Issue #2771, opened 2018-10-18
Refactor testing to include new e2e tests and make it easier to run phpunit locally


Shelob9:
Copy e2e tests and Docker from https://github.com/CalderaWP/calderawp-wordpress-plugin/tree/master/cypress

- [x] Scripts to install and start local dev with docker.
- [x]  Scripts to run e2e tests with docker environment.
- [x]  Scripts/setup for actual unit tests
- [x] Scripts to run existing integration tests with docker. Should not break how Travis runs scripts.
- [x] Run e2e tests in Docker container in Travis

Related #2765

Shelob9:
close via 23e4982

PR #2767, opened 2018-10-17
Test JS tests in Travis

New0:

PR #2772, opened 2018-10-18
Refactor testing

Shelob9:
#2771 

This patch makes the following changes:

- Adds isolated PHP unit testing, separate from existing phpunit integration tests that use the WordPress test suite.
- Add e2e tests powered by Cypress. We can run these locally and trigger builds to record in Cypress.io. This is not setup in CI yet. https://github.com/CalderaWP/Caldera-Forms/pull/2772#issuecomment-431388047
- Adds a local development environment with all of the build and test steps scripted.
- Allows WordPress integration tests to run the traditional way -- as they will continue to run in Travis -- or in Docker.

We also have JS unit tests running in CI now #2767


Shelob9:
@New0  This is now working locally. No CI yet (or xdebug or HTTPS).

Note: Caldera Forms now has its own dev environment. So instead of checking this out to `wp-content/plugins` keep it separate from existing site. 

Please setup local dev according to instructions here:
https://github.com/CalderaWP/Caldera-Forms/blob/9429337f39203179a7d4805b4048bc23abd1f19b/README.md#install-test-environment

Make sure WordPress works and the test forms and pages were created by the last command. This will confirm this works on Windows.

Review the testing structure (we now have PHP integration, PHP unit, JavaScript unit and e2e).
https://github.com/CalderaWP/Caldera-Forms/blob/9429337f39203179a7d4805b4048bc23abd1f19b/README.md#test-structures

Make sure you can run the each of the 4 types of tests:
- SEE commands: https://github.com/CalderaWP/Caldera-Forms/blob/9429337f39203179a7d4805b4048bc23abd1f19b/README.md#commands
- Run php unit tests (I added one example test to new test suite in `tests/Unit`
- Run JavaScript unit tests (with Jest)
- Run php integration tests (pre-existing phpunit tests)
- Run new e2e tests

Shelob9:
#2485

Shelob9:
I reverted a068018 which set up Cypress recording in Travis (attempted to anyway.) I don't want that on every push and I don't want to rewrite Travis config again and make it way more complex than I can grok. This would be perfect for Github actions, I'll wait until we get added to the beta, I requested access.

I will merge this, if it works on @New0 windows computer.

New0:
First try resulted : 
- wp:install failed in the last command of install-docker.sh l67
  - I tried with bash ./bin/install-wp-tests.sh instead of /app/bin/install-wp-tests.sh and it got called but failed l23 syntax error near unexpected token `elif' and got error with line breaks and not finding l4 ./bin/includes.sh

swapnilmanjrekar:
@New0 #2772 Regarding your " wp:install failed in the last command of install-docker.sh l67" error, I ran the commands for CF local on my windows command line terminal , I am using windows 10 pro os and windows command line terminal for running the command instead of gitbash! I suggest you delete git from root directory of your drive and install GIT and choose Windows Command Line Terminal while installing it !
And then run the following commands:

1.  git clone git@github.com:CalderaWP/Caldera-Forms.git
2.  cd Caldera-Forms
3.  git fetch feature/2771
4.  git checkout feature/2771
5.  git pull origin feature/2771
6.  composer install && npm _update_
7.  composer wp:install
8.  composer wp:activate
9.  composer test:unit

Shelob9:
Tests not working in PHP5, beacuse the 2nd argument for dirname() -- levels -- was added in php7. Fuckers.

New0:
@Shelob9 , I realize this PR is merged and closed but in case I need to remember how I installed local dev testing environment I'll leave notes here anyway.

here are the two steps I took to get composer dev:install install correctly the environment :

- Edited l67 in bin/install-docker.sh "/app/bin/install-wp-tests.sh" part, to "bash $(dirname "$0")/install-wp-tests.sh"

( The whole line 67 became :
docker-compose run --rm wordpress_phpunit bash $(dirname "$0")/install-wp-tests.sh wordpress_test root example mysql "${WP_VERSION}" false >/dev/null
)

- I used Git bash and ran :
sed -i 's/\r$//' bin/install-docker.sh
sed -i 's/\r$//' bin/includes.sh
sed -i 's/\r$//' bin/install-wp-tests.sh

- then I could run composer:dev:install correctly all the way

Issue #2766, opened 2018-10-16
Integrate Adv File Fields 2.0 in Caldera Forms


Shelob9:
- [ ] Install dependency (https://github.com/CalderaWP/caldera-components/pull/18)
- [x] Field type registration
- [ ] Settings (should be same as existing adv file)
- [x] Handler function
- [x] Front-end view, empty div for Render app

Changes to front-end rendering to accommodate React-rendered fields such as this.
- [x] React component - `CalderaFormsForm` use portals to render part of form markup, leave rest alone.
    - https://github.com/CalderaWP/Caldera-Forms/blob/master/assets/js/frontend-script-init.js#L349
- [x] In init function, init a React app (no React DOM). Inject state
    - https://github.com/CalderaWP/Caldera-Forms/blob/master/assets/js/frontend-script-init.js#L349
- [x] Use CF Events to subscribe to changes and pass to React app.
- [x] Use CF Events to send event based on result of switch in conditionals.js'
    - https://github.com/CalderaWP/Caldera-Forms/blob/master/assets/js/conditionals.js#L219-L266
    - Subscribe to changes to update hide/show state
    - Need a way to supply value to be evaluated to conditionals.

Shelob9:
Close via  954c14e #2775

Issue #1399, opened 2017-03-14
Option to limit file size for adv file 2.0 and standard file fields


TheGreenArrow:
The file size control is working by file instead of the total files size.

![2017-03-14 11 31 06](https://cloud.githubusercontent.com/assets/21695752/23896511/fd257162-08a9-11e7-928f-55a6317bc8f7.jpg)

In this example, limit of file size uploading is 2Mo.

Shelob9:
Correct, that is how this works. I'm going to retitled this as a feature request. As it would be a cool feature if it could do that.

Shelob9:
Changing name to reflect that is will be implimented in standard file file fields and advanced file fields 2.0, not advanced file fields 1.0.

Shelob9:
Close via 6e707fd #2823

PR #2775, opened 2018-10-18
[WIP] Integrate new file fields

Shelob9:
#2766


Shelob9:
NOTES to resolve before merge:

- [ ] I hardcoded `    add_filter( 'caldera_forms_render_assets_minify', '__return_false' );` in main plugin file. Dev environment needs a mu-plugins dir for this.
- [ ] I removed init of old trackign system, which we should do.
- [ ] I removed freemius, which we should not do

Shelob9:
To finish this:

1) @New0 Make button look like the old button https://github.com/CalderaWP/Caldera-Forms/blob/9ecc5d67b9d5426516ce2b1f2dd399cd0b8f5f5e/clients/tests/unit/render/Fields/FileInput.test.js
2) @Shelob9 Make cf2 fields repeat
3) @Shelob9 Finish making it submit and write tests for that.

Shelob9:
@New0 For button style:
1) It should look the same (the drag or click to upload element) as an existing file field. This should be done by applying existing rules using same classes. IE If I had an advanced file field and targetted it with custom CSS in my theme, those rules should have same effect on new file field. Be careful that `cf-uploader-trigger` class is what advanced file fields bind to, and using that is a bad idea.
2) Drag over should still work.

Shelob9:
TODOS

Front-end @New0
* Style 
* Image preview (this is a feature I will let go off)
* Support multiple
* Submit

Server-side @Shelob9 
* Integrate with existing abstractions for upload handler, file path.
* Tests for above
* Test form submission with cf_file fields

* e2e Tests @Shelob9 
* Integration tests for wp_mail (attached properly)
* Integration tests for CF Pro emails (attached properly, there is an API endpoint for this)


screen shot 2018-10-25 at 10 54 04 am

clients/render/components/Fields/FileInput.js
```diff
@@ -26,22 +26,24 @@ export const FileInput = (props) => {
 		});
```
Shelob9: This is for #2766 

@New0 All commits must have an issue number.

clients/render/components/Fields/FileInput.js
```diff
@@ -84,15 +84,8 @@ FileInput.defaultProps = {
 	multiple: true,
```
Shelob9: For #2766

clients/render/components/Fields/FileInput.js
```diff
@@ -2,35 +2,50 @@ import {CalderaFormsFieldGroup, Fragment} from "../CalderaFormsFieldGroup";
...
-	} = field;
-
-	const onDrop = (acceptedFiles, rejectedFiles) => {
-		acceptedFiles.forEach(file => {
-			onChange(file);
+export class FileInput extends React.Component {
 
-		});
-	};
+  constructor(props) {
```
Shelob9: @New0 Why use a stateful component here?

The disadvantage that I see is that the File object is now stored in two places -- this component's state and the form render component's state.

I'm not sure what the advantage is. I see that you are using `this.state.files` to build the preview, I do not see an advantage of this versus itterating on `props.fieldValue` which should be one File object or an array of File objects.

clients/render/components/Fields/FileInput.js
```diff
@@ -22,7 +22,17 @@ export class FileInput extends React.Component {
         files: [...prevState.files, file]
       }))
 		})
-
```
Shelob9: For #2766

Shelob9:
@New0 Summary of my comments on your commits:
1) `FileInput`'s sole responsibility is the interface, and therefore should NOT be a stateful component. The business logic and state management for ALL components that are children of `CalderaFormsRender` should be contained in `CalderaFormsRender`. 

2) None of this code has tests.  You should be writing a test to prove the change you want to make works, commit the failing test and then committing a change that makes the tests pass.

Shelob9:
@New0 The functionality of the `FileInput` and look is much closer, good work. 

You should have a prop for if multiple files are allowed or not. Make sure to only show the add more if that is true.

Also, make sure that you do not try and preview non-images:
screen shot 2018-10-26 at 12 45 49 pm

Shelob9:
Summary of how this should work:

When field is added in the FileInput, the File object gets set in state of the main component.

When form submit, we loop through all fields, if they are a file field, we read the file object to calculate the hash:
https://github.com/CalderaWP/Caldera-Forms/blob/19283f6d4644f052d9515ff812d1f1ec685935d0/clients/render/index.js#L95

When that's done, the file is submitted against a new REST API endpoint
https://github.com/CalderaWP/Caldera-Forms/blob/19283f6d4644f052d9515ff812d1f1ec685935d0/clients/render/index.js#L60

https://github.com/CalderaWP/Caldera-Forms/blob/19283f6d4644f052d9515ff812d1f1ec685935d0/cf2/RestApi/File/CreateFile.php#L87 <- BTW the file hash check is bypassed because that hash generation I showed earlier and my server-side hashing do not agree yet.

That does the upload: https://github.com/CalderaWP/Caldera-Forms/blob/19283f6d4644f052d9515ff812d1f1ec685935d0/cf2/Fields/Handlers/FileUpload.php#L40

And sets the data in a "transient" https://github.com/CalderaWP/Caldera-Forms/blob/19283f6d4644f052d9515ff812d1f1ec685935d0/cf2/RestApi/File/CreateFile.php#L122

This is the same basic way it works for existing adv file fields. One or more sessions sets transient, then we exchange transient ID for transient value when field is processed during main submission.

https://github.com/CalderaWP/Caldera-Forms/blob/feature/2766/cf2/Fields/Handlers/FileFieldHandler.php#L30

Shelob9:
@New0 I wrote pseudo-e2e tests here:
https://github.com/CalderaWP/Caldera-Forms/blob/9056df97f84aaae871a5cd3a929f631e1cf9c2ed/cypress/integration/cf2-file-ui.js

Please make sure as you're working on the front-end, that you consider all of those tests. Not sure if I will have time to write them for real before Tuesday and if it would make sense to write these tests first (TDD is for unit tests, I think.) But, that describes the behaviour we want (make sure to check me.)

Shelob9:
Markup for remove button would be roughly like this:

`<button data-file="fld_5802220_1_file_fl51515810816029" class="cf-file-remove"><span class="screen-reader-text sr-text">Click To Remove</span></button>`

I don't think it should look like a button, I think it should look like a little red X like the old version. No one ever complained about that. As long as the default CSS is easy to ovveride, 🌮 .

Shelob9:
#a11y concerns:

Right now, this is screenshot of field with 2 files added and its HTML.
screen shot 2018-10-30 at 10 42 53 am

- Right now the id attribute is on the outter div, not the file input, making the  label's for attribute confusing, I think. We can pass the ID to the input instead. 
- We already discussed the remove button above.
- The file list. How do we relate that list to that input.
- The add files button, its a button element, it describes in text what it does. That's what we need right?

Shelob9:
RE: accesibility and the hidden input/button: https://developer.mozilla.org/en-US/docs/Web/API/File/Using_files_from_web_applications

Shelob9:
After 32d6ff7 the uploads works, if I slow it down with xdebug. IE we do not have the transient value set yet when it submits, so its not there when the email happens. I think.

assets/js/ajax-core.js
```diff
@@ -129,22 +129,43 @@ jQuery(function($){
...
+                 *
+                 * @since 1.8.0
+				 */
+				$( document ).trigger( 'cf.form.request', {
+                    $form: $form,
+                    formIdAttr: $form.attr( 'id' )
+                });
+
+				//Check if any cf2 fields are blocking submit
+				var cf2 = window.cf2[ $form.attr( 'id' ) ];
```
Shelob9: @New0 I shoved it into a window scoped object I was already using. That's better than a window scope object, I think.

assets/js/ajax-core.js
```diff
@@ -129,22 +129,43 @@ jQuery(function($){
...
+
+				/**
+                 * This event is triggered directly before the request for form submission is made
+                 *
+                 * Runs after cf.form.submit
+                 *
+                 * @since 1.8.0
+				 */
+				$( document ).trigger( 'cf.form.request', {
+                    $form: $form,
```
Shelob9: @New0 Should we be sharing phpStorm settings? Not sure mine are to blame or yours, but this is not optimal.

 We need eslint or prettier #2485

Shelob9:
Submitting, in Safari and attaching to emails after e405c72efca1458b279a12785f0c7ebd252fa46a 🌋 🤘

assets/js/ajax-core.js
New0: Absolutely @Shelob9, do you to share your code style settings ?
https://www.jetbrains.com/help/idea/copying-code-style-settings.html

Shelob9: LOL I have that post in my list of things to do. I'll try it out today.

clients/render/components/Fields/FileInput.js
```diff
@@ -0,0 +1,197 @@
...
+  const removeFile = (e, file) => {
+
+    const index = fieldValue.indexOf(file);
+    fieldValue.splice(index, 1);
+
+    onChange( fieldValue );
+
+  }
+
+  let ulExpanded;
```
Shelob9: @New0 This naming is a little confusing. `isListExpanded` is more obvious.

Also, make it one line - `const isListExpanded = fieldValue.length > 0;`

cypress/integration/cf2-file-ui.js
```diff
@@ -0,0 +1,129 @@
...
+describe('Name of test', () => {
+	beforeEach(() => {
+		visitPage('hello-world');
+	});
+
+	const formId = 'cf111';
+
+
+
+	it( 'Can add an image and see preview', () => {
```
Shelob9: @New0 Use Jest and/or Enzyme to do a test with same spec directly on `FileInput`.

Shelob9: Actually all of these tests should be possible using Jest and/ Or Enzyme

Shelob9: @New0  BTW I added Enzyme and used it here: https://github.com/CalderaWP/Caldera-Forms/blob/feature/2766/clients/tests/unit/render/RenderComponentViaPortal.test.js

Shelob9:
@New0  With 9eafdd7 the prop for accept is '.png' or '.png,.jpg' which I thaught would work. It needs to be `image/png`. I'll fix that server-side.

I hard coded it to `image/png` then tried to upload a jpg, it didn't accept it, but it didn't show an error. We need to use the reject callback to add messages. That callback will need to be in scope of `CalderaFormsRender` as the error message it generates would go into `CalderaFormsRender.state.messages[fieldIdAttr]`;

clients/tests/unit/render/Fields/Mocks/MockFileFieldRenderer.js
```diff
@@ -0,0 +1,65 @@
+import {FileInput} from "../../../../../render/components/Fields/FileInput";
+import React from 'react';
+export class MockFileFieldRenderer extends React.Component {
```
Shelob9: @New0 This is a fake renderer I am mounting the file input in so we can have state one level above it without dealing with the portal render.

clients/tests/unit/render/Fields/Mocks/MockFileRenderer.test.js
```diff
@@ -0,0 +1,78 @@
...
+		component.instance().setMessage(message );
+		expect(component.state('message')).toBe(message)
+	});
+
+});
+
+describe( 'DOM testing file components', () => {
+
+
+	it( 'If multiple upload option is not enabled, can not add a second file', () => {
```
Shelob9: @New0 Here is an example test. Doesn't even require a file blob. I'm just checking based on number of values.

We need to use file blobs for testing the preview.

For the onDrop/onReject, we should use file blob to test that blobs get passed to array properly. We do not need to test that dropzone knows how to reject/accept files.

We need to test that when a message that indicates an error is passed to  the field that it shows correctly.

Shelob9:
@New0 I looked at your test fixture. it was not causing the `setFieldValue` method in the container component to be called. I added a data attribute to the wrapper, which it should have had, and then update the util function so it targets the dropzone. https://github.com/CalderaWP/Caldera-Forms/pull/2775/commits/323d96cad5c1c0ed1696356af54a85f52fc90746#diff-34b03f549644946ea4fe9528e482efc2R81

Now we are receiving an empty array at the `setFieldValue` method in the container component to be called. So either your fixture isn't passing the file object right (I don't think so) or this isn't possible.

In cb0f428 I added a failing test with Enzyme to show that having a value passed to a field that should not accept multiple values hides the add more button. I think we can get the coverage we need (this basically what I said in slack) with that and isolated testing of the onChange()/ onDrop(). 

As we discussed you finish testing, then work on the button (accessible X icon) and #a11y  concerns.

cf2/Fields/RenderField.php
```diff
@@ -130,6 +130,7 @@ public function data()
                 'multiUploadText' => ! empty($this->field[ 'config' ]['multi_upload_text' ]) ? $this->field[ 'config' ]['multi_upload_text' ] : false,
                 'allowedTypes' => $this->getAllowedTypes(),
                 'control' => uniqid($this->getOuterIdAttr() ),
+                'usePreviews' => ! empty($this->field[ 'config']['use_image_previews']) ? $this->field[ 'config']['use_image_previews'] : false,
```
Shelob9: @New0 Small note - only valid values we want to pass to DOM are true and false, but what if `$this->field[ 'config']['use_image_previews']` is `pants` or `['foo' => 'bar' ]` ?

`! empty($this->field[ 'config']['use_image_previews']) ? true : false` removes a whole class of edge cases.

Shelob9:
@New0  When we talk next, let's discuss naming, labeling, icon and if we move existing advanced file fields to deprecated or what not.

Shelob9:
@New0 I am going to merge this to develop and tag an alpha today.  This works. #2814 and #1399 can be separate issues we can get into alpha 2.

PR #2779, opened 2018-10-20
Import calculation values properly

### LINKED ISSUE #2778: Calculation value of field options is always zero after import
IF I have a field with calc values:
screen shot 2018-10-20 at 12 44 06 pm

After it imports the calc_values are 1.
screen shot 2018-10-20 at 12 44 40 pm

This is bad and also blocking #2772 because none of my caluculations tests pass for stupid reasons.

Shelob9:
Fixes #2778 

To test, create a form with a select field that has calculation fields that are not 1.

Then export and import the form, make sure the calculation values imported correctly.


Shelob9:
With 959eeb2  we can now get the calc value passed from value or label, like it works in front-end if its not there. With out that, calc value became 0, if no calc value was in export. But if label was 7, no calc value set, original form calculates 7 for that option and imported form calculates 0, which is bad. With 959eeb2 that option calculates 7, which is good.

PR #2794, opened 2018-10-26
Private persistent file upload

pascalknecht:
This PR aims to give the user the possibility to store the files on the server if he does not want to have it deleted.


Shelob9:
@pascalknecht Thanks for the pull request. We will consider this after #2775 is merged.

Can you please open an issue explaining the reason for why you would like this change to be made?

pascalknecht:
@Shelob9 I created the issue you wanted #2804. Please keep me up-to-date with the changes.

fields/advanced_file/config_template.php
```diff
@@ -1,61 +1,70 @@
-<div class="caldera-config-group">
```
Shelob9: @pascalknecht  I see what you are doing with the role attribute. The problem here is that this patch is about adding 1 setting. By modifying every setting its going to make merging it with changes in #2766 

Please revert changes to existing fields. This is a legit #a11y issue that we will not repeat when we refactor this UI.

classes/files.php
```diff
@@ -226,6 +226,8 @@ public static function cleanup( $form, $second_run = false ){
         $fields = Caldera_Forms_Forms::get_fields( $form, false );
         foreach( $fields as $id => $field ){
             if( Caldera_Forms_Field_Util::is_file_field( $field, $form ) ){
+                if ((bool)$field['config']['persistent']) return;
```
Shelob9: Please use braces around conditionals.

fields/advanced_file/config_template.php
Shelob9: Hold on, I'm wrong. Looks like this diff is beacuse of auto-formatting. 

Either way we need it to just be adding that 1 setting

Shelob9:
@pascalknecht Thanks for the change. Here are some notes. Once #2775 is merged, this is will likely need some work to align it there. 

 Please see the note I left on the diff.

Also:
1) This PR is going to need to support the new file field #2776 
2) I think all that means is using the new setting for persistent  to the config file for new file field https://github.com/CalderaWP/Caldera-Forms/blob/feature/2766/fields/cf2_file/config.php

fields/advanced_file/config_template.php
pascalknecht: @Shelob9 Yes they are indeed only auto formatting changes. Tried to do that but it always inserts those formats.

pascalknecht:
@Shelob9 Can I do another PR for #2776 as this will need a merge between my branch and develo?

Shelob9 (commented):
I think this is both good to go and blocked by #2819 

@New0 Let's address if this should be a part of 1.8 or wait until 1.8.1

PR #2802, opened 2018-11-06
Create a proper job manager for deleting transients etc.

### LINKED ISSUE #2801: Use a proper job manager for our "transients" API
In #1866 I moved all of our "transients" to a wrapper around the options API and reproduced the delte option after X time API that the real transients API in WordPress offers, but is not guaranteed. This solved the situation where an object cache is in use, but memory is maxed out (memcached is 1GB max) and so the file uploads in one session and then the transient was gone by the next session.

Clearing "cron" jobs has never worked well, leaving to orphaned options. I wrote a prototype and Torque article for using a5hleyrich/wp-queue as a job manager.

Goals:
1) Use wp-queue to schedule deletes of our transients.
2) Add a feature to CF Pro that pings an endpoint to run queue so it doesn't have to use wp "cron"
3) Add a feature to CF Pro for delayed auto-responders and such.

Shelob9:
Resolves #2801 

This PR introduces:

1) A proper job scheduler that we can run ourselves or with wp-cron.
2) Service registration on container.

Also, it uses new job manager to schedule deletes of transients. By default it is database-backed and runs on wp_cron. I will provide a REST API endpoints, secured with CF Pro API keys once I think that through, that can be its own PR.


New0:
Rebuild PR for Travis tests

New0:
Rebuild for Travis tests

New0:
Re-run tests

New0:
Re-run Travis tests

New0:
Fixed #2801

PR #2808, opened 2018-11-07
Allow validation to be re-triggered on the phone field

### LINKED ISSUE #2727: Date Picker Validation not updating on selecting date
When using Date Picker field as required, when we submit the form validation works as indented, applying error class and showing message, but like other fields the validation is not updating when we select date, eg: for text field when we start typing, error validation will be removed, so for Date Picker filed It should be updating the validation on select, but only after submitting again the form validation gets updated, to better show the scenario please see the attached image

caldera-issue

Wordpress: 4.9.8
Caldera Forms: 1.7.2
PHP: 5.6.31

Shelob9:
Resolves  #2727 by setting the data-cfdatepicker attribute for date picker fields so blur event is triggered #2727

See https://github.com/CalderaWP/Caldera-Forms/issues/2727#issuecomment-436744891


New0:
@Shelob9 could you push the form used for the cypress test ? (date-validation-2747.js)
It's Form ID CF5be31da4beff7 (.json) to cypress/forms

(My manual test didn't pass, clicking datepicker field didn't open the datepicker )

Shelob9:
screen shot 2018-11-15 at 9 17 35 pm
Holy shit, I do.

Issue #2821, opened 2018-11-21
Uncaught TypeError when submitting form with no cf2 fields.


Shelob9:
![screen shot 2018-11-21 at 5 09 40 pm](https://user-images.githubusercontent.com/1994311/48870825-473ce400-edb0-11e8-90d2-a726ef63e2ae.png)

When I submit a form, that has no cf2 file fields, I get this error: `Uncaught TypeError: Cannot read property 'CF5bf3166fbca21_1' of undefined` It  happens because `window.cf2` never got set because we didn't need it really, but now, we need to call it.

Shelob9:
Close via e8dae2e #2822

Issue #2820, opened 2018-11-21
Use output buffering to prevent PHP warnings/notices from returning bad JSON


Shelob9:
Would prevent issues like #2448 and #2788

Shelob9:
Will be fixed via #2822

Issue #2788, opened 2018-10-24
The page is not refreshed when editing the entries


evolluti0n:
## Do You Need Immediate Help?
* A ticket is already open
https://secure.helpscout.net/conversation/687067564/13287?folderId=2372292

## Is This A Feature Request?
* No. It's not a feature request.

## What Version Of Caldera Forms, WordPress and PHP Are You Using?
* PHP 7.2.11
* WordPress 4.9.8
* Caldera Forms 1.7.4

## Does Your Issue Persist When You Disable All Other Plugins and Switch To The Default Theme?
Yes. The issue persist with all plugins disabled and with the default WordPress theme (Twenty seventeen)

## What Is The Unexpected Behaviour?
I enabled the option to create a sub-menu entry viewer. This add a new link in the left sidebar in the WordPress dashboard in Caldera Forms group. If I edit entries from this link the error happens. Every time that I edited the entries for any form with this option enabled, no matters the field, and click save, a loading icon starts to spin but never ends. The information is actually updated in the form, but the page is not refreshed.
If you select the entries from Caldera Forms > Forms > form name > Entries > View > Edit, the error doesn't occur.​​ I enabled the ​​same option for other forms that I have in my installation, and got the same issue.

## What PHP Errors Have You Logged While Reproducing This Bug?
16:45:13.331 vue.min.js:2 [Vue warn]: Cannot find element: #caldera-forms-clippy
te @ vue.min.js:2
Xa @ vue.min.js:1
Ka.$mount @ vue.min.js:3
Fa @ vue.min.js:1
a._init @ vue.min.js:2
Ka @ vue.min.js:1
(anonymous) @ caldera-clippy.js:45
(anonymous) @ load-scripts.php?c=1&load[]=jquery-core,jquery-migrate,utils,jquery-ui-widget,jquery-ui-position,wp-pointer,zxcvbn-async,password-strength-meter,jquery-ui-c&load[]=ore,jquery-ui-tabs,jquery-form&ver=4.9.8:2
i @ load-scripts.php?c=1&load[]=jquery-core,jquery-migrate,utils,jquery-ui-widget,jquery-ui-position,wp-pointer,zxcvbn-async,password-strength-meter,jquery-ui-c&load[]=ore,jquery-ui-tabs,jquery-form&ver=4.9.8:2
[...]

PR #2822, opened 2018-11-21
Use output buffering to prevent server-side errors form printing in json response

Shelob9:
Changes made:

* Fix window.cf2 not being an object #2821
* Start output buffering when form submission starts. #2820 #2448 #2788

Not going to  use for admin-ajax usage in admin beacuse then I have to worry about testing that and we're going to rip that all out soon anyway. The entry edit bug  is covered by this PR without doing any of the other admin send jsons.


Shelob9:
I'm not sure how to test this with automation, but from manually testing I know that if this is wrong, it will break every acceptance test that actually submits the form, so that's a good test.

I added this filter locally:

```
add_action( 'caldera_forms_submit_process_before', function() {
	echo 'aaaa';
	echo $x;
});
```

Before this patch, ajax response would be like `aaaa. NOTICE $x not defined {data: ...}` goal is to only return JSON.

Shelob9:
@New0 Please review and merge if good.

regiside:
I'm running into this #2448 issue. Is this change available for download yet? If not when/how can I get this into my version of Wordpress. Thanks for the help!

Shelob9:
@regiside This will be included in Caldera Forms 1.8. We will release a beta of 1.8 from the develop branch, once it is stable. 1.8 is a major update and it's not ready for use on a real site.

regiside:
Okay thanks, for the time being I've just disabled the Ajax functionality as I was redirecting after form submission anyway. I'll look forward to 1.8!

Issue #2818, opened 2018-11-21
cf2 file fields, only one file is attached and uploaded


Shelob9:
@New0  found that for files field, when multiple is enabled, only one file is  attached to mailer and one file uploaded to media library.

Shelob9:
Relateded sort of: #2814

Shelob9:
1) There is a typo in the inline docs about what type of argument for function `hashAndUpload` in render/index.js

2) I observed that if I have one file field and I add two files, when I submit, I only see 2 HTTP requests, I would expect three. In the media library I see that only the first was added, I would expect to see two.

The tests here, which are server-side tests prove that if the right data was already on the server, it would attach both.
* https://github.com/CalderaWP/Caldera-Forms/blob/develop/tests/Integration/SubmissionsWithFileFieldsTest.php
* https://github.com/CalderaWP/Caldera-Forms/blob/develop/tests/Integration/FileUploadTest.php

I think the problem is purely in the client, only one upload happens before allowing submission to complete.

Shelob9:
After uploading first file ~L155 clients/render/index.js  the field is removed from pending, without triggering upload of second file.
```
else if (response.hasOwnProperty('control')) {
									removeFromPending(fieldId);
									removeFromBlocking(fieldId);
									cf2.uploadCompleted.push(fieldId);
									$form.submit();
								}else{```

Shelob9:
To make this testable, we need to move the closure for `cf.ajax.request` a function. Then we need to make the response handlers separate functions. Then we can test that if there are 2 files, only gets removed from cf2.pending if both are uploaded. We need to test that the cf2 object is updated correctly. 

Given that file field has 2 files. Expect one file uploaded and submit is still blocked based on that field. Given that second file has uploaded, expect submission not to be blocked.

Mocking jQuery - https://stackoverflow.com/a/45997857/1469799

Pseudo-tests https://gist.github.com/Shelob9/0ed44838d53e3aacd349136f5a6d7135#file-pseudo-tests-js

New0:
Fixed by #2910

PR #2823, opened 2018-11-22
 Limit file upload size option

Shelob9:
#1399 

I merged `feature/2814` (PR #2819). That PR should get merged first probably. 

This adds a UI option for max upload size for HTML5 file  fields and adv file fields 2.0.

This checks server-side if adv file fields 2.0 size is too large and rejects.

We might wish to, in order of how much I care:
1) Check server side on HTML5 file fields. Important or need to revert UI option.
2) Check in the broweser for file size of adv file field 2.0. Important, but should be an issue in 1.8.1 milestone.
3) Show max file size in error message. Not sure if I like this, the file field description can be used to indicate max size.
4) Implement in adv file 1.0. I disagree with doing this. I'm done maintaining that.


New0:
When a Max upload size is set the error "SyntaxError: "JSON.parse: unexpected character at line 1 column 1 of the JSON data" prevents the submission. It is an error that will be fixed by the index.js refactor, this PR is blocked by issue #2818

Shelob9:
@new0 That sounds like a PHP error during submission. Please retest with PHP debug logging on and report what the error is or what the lack of error is. Please check for PHP errors when testing.

Issue #2840, opened 2018-12-03
jQuery load function error


TarikTz:
## What Version Of Caldera Forms, WordPress and PHP Are You Using?
WordPress Version: 4.9.8PHP Version: 7.0.32-0ubuntu0.16.04.1MySQL Version: 5.7.24Caldera Forms Version: 1.7.4WP_DEBUG: 1

## Does Your Issue Persist When You Disable All Other Plugins and Switch To The Default Theme?
Yes

## What Is The Unexpected Behaviour?
Conflict with new version of jQuery, .load function is depricated. 

## What JavaScript Errors Have You Seen While Reproducing This Bug?
`Uncaught TypeError: a.indexOf is not a function`

Shelob9:
By new version do you mean jQuery 3.0?

We intentionally target the version of jQuery that ships with core and do not support any other version. 

I would consider this if if is 100% backwards compatible with the correct version (WordPress' version) but I'd say I'll need to be convinced as I don't want to get into testing and supporting other versions.

TarikTz:
> By new version do you mean jQuery 3.0?
> 
> We intentionally target the version of jQuery that ships with core and do not support any other version.
> 
> I would consider this if if is 100% backwards compatible with the correct version (WordPress' version) but I'd say I'll need to be convinced as I don't want to get into testing and supporting other versions.

I'm using Wordpress with the latest version of Bootstrap and jQuery 3+ so for my project Caldera form was throwing me errors when trying to submit the form. I can stick with custom fixes when the new versions are released, but I would like if there would be some kind of an option for us using the latest versions of the script.

PR #2841, opened 2018-12-03
Fixed .load function problem

TarikTz:
Fixed .load function with .on('load',...) - .load is deprecated in new jquery


Shelob9:
For #2840 

@TarikTz thanks for the PR. Note that all pull requests must be submitted against the develop branch.

Issue #2843, opened 2018-12-04
Fatal error when requesting forms via API with ?full=true&details=true 


Shelob9:
Related(ish) #2692

GET Request to:
`http://localhost:8218/wp-json/cf-api/v2/forms?full=true&details=true`

Error message is:
```
<b>Fatal error</b>:  Uncaught TypeError: Argument 1 passed to Caldera_Forms_API_Forms::prepare_form_for_response() must be an instance of Caldera_Forms_API_Form, null given, called in /var/www/html/wp-content/plugins/caldera-forms/classes/api/forms.php on line 270 and defined in /var/www/html/wp-content/plugins/caldera-forms/classes/api/forms.php:382
Stack trace:
#0 /var/www/html/wp-content/plugins/caldera-forms/classes/api/forms.php(270): Caldera_Forms_API_Forms-&gt;prepare_form_for_response(NULL, Object(WP_REST_Request))
#1 /var/www/html/wp-includes/rest-api/class-wp-rest-server.php(936): Caldera_Forms_API_Forms-&gt;get_items(Object(WP_REST_Request))
#2 /var/www/html/wp-includes/rest-api/class-wp-rest-server.php(321): WP_REST_Server-&gt;dispatch(Object(WP_REST_Request))
#3 /var/www/html/wp-includes/rest-api.php(266): WP_REST_Server-&gt;serve_request('/cf-api/v2/form...')
#4 /var/www/html/wp-includes/class-wp-hook.php(286): rest_api_loaded(Object(WP))
#5 /var/www/html/wp-includes/class-wp-hook.php(310): WP_Hook-&gt;apply_filters('', Array)
#6  in <b>/var/www/html/wp-content/plugins/caldera-forms/classes/api/forms.php</b> on line <b>382</b><br />
```

I know how to fix.

PR #2844, opened 2018-12-04
Use property with form object 

Shelob9:
Resolves #2843 

The method `form_object_factory` by design (questionable design by past Josh) sets the class property form and returns nothing, which was causing the error. So I'm passing $this->form on the line that was previously giving me the error.

PR #2845, opened 2018-12-04
Prevent fatal error on REST API request for forms with full & details args set to true

Shelob9:
For #2843 

Changes: 

- The method form_object_factory by design (questionable design by past Josh) sets the class property form and returns nothing, which was causing the error. So I'm passing $this->form on the line that was previously giving me the error. 92c7191

- Tests that produces the same error, before fix. ca02cc8


Shelob9:
This PR replaces #2844

After ca02cc8 test makes the same error:

screen shot 2018-12-04 at 11 50 20 am

Shelob9:
@New0  This is now ready for review.

Issue #2849, opened 2018-12-05
Custom autopopulated checkboxes are not calculated by Calculation field


mecachisenros:
## Do You Need Immediate Help?
Nope

## Is This A Feature Request?
More of a fix and feature request.

## What Version Of Caldera Forms, WordPress and PHP Are You Using?
WordPress Version: 4.9.8
PHP Version: 7.1.24
MySQL Version: 5.7.24
Caldera Forms Version: 1.7.4

## Does Your Issue Persist When You Disable All Other Plugins and Switch To The Default Theme?
Yes :)
## What Is The Unexpected Behaviour?
When using `caldera_forms_field_option_presets` and `caldera_forms_render_get_field` hooks for autopopulate/bulk insert presets options, is common practice to use some unique ID (an `integer`) for the populated data.

When using those unique IDs and also using the field for autopopulation, Checkbox fields are ignored because `Caldera_Forms_Field_Util::get_option_calculation_value()` checks very early for `is_string()` [here](https://github.com/CalderaWP/Caldera-Forms/blob/2cdb95b7f258b6567f0b570da526ab4bd6d58ec3/classes/field/util.php#L523) and `caldera_forms_get_option_calculation_value` hook never kicks in because of the [early return](https://github.com/CalderaWP/Caldera-Forms/blob/2cdb95b7f258b6567f0b570da526ab4bd6d58ec3/classes/field/util.php#L530) that evaluates the calculated value to `0`. `is_string()` makes sense for Caldera Forms' implementation of options as those get an ID in the form of `optXXXX`, but for custom implementation I think it would be appropriate to also accept an integer as the option ID.

So that early check could also check for `is_int()` or `is_numeric()` or both, thoughts?
```
...
if ( is_array( $option ) || is_int( $option ) ) {
...
```
Happy to submit a PR if agreed.

## What PHP Errors Have You Logged While Reproducing This Bug?
No errors to reproduce.

## What JavaScript Errors Have You Seen While Reproducing This Bug?
No errors to reproduce.

Shelob9:
Thanks for reporting. Your idea for how to fix this makes sense.

Am I correct that this happens because you've chosen to make the option IDs the same integer as the value when loading the forms via JSON export or the filter for file system-based forms? Makes logical sense to do that, we're just trying to make sure we understand root of the bug.

mecachisenros:
[...]

PR #2850, opened 2018-12-05
Allow integers as option ID for custom autopopulated checkboxes

mecachisenros:
As per my description in #2849 this will allow custom autopopulated Checkbox fields to be calculated when used as an Operator in a Calculation field.


Shelob9:
@mecachisenros  Thanks for the PR. We need  to add one test before this can be merged.

Test will need a checkbox field with options like this:

`[ config => options => [ 'opt12345' => [ 'value' => 1 ],  2 => => [ 'value' => 2 ] ]`

Then it will test
* Given that form with an option like the one above -- one option with a numeric option ID of the string `opt12345` and an option ID of integer `2`, when calling `Caldera_Forms_Field_Util ::get_option_calculation_value()` 

* assert that the filter `caldera_forms_get_option_calculation_value ` is called twice.

mecachisenros:
@Shelob9 thank you.

Could you point me to the test file where I should write the test?

Shelob9:
@mecachisenros Tests would be written here: https://github.com/CalderaWP/Caldera-Forms/tree/develop/tests

Shelob9:
@mecachisenros I should clarify, php unit tests for the function you modified would go there. That doesn't cover the front-end JS preview. We would need a cypress test for that - https://github.com/CalderaWP/Caldera-Forms/tree/develop/cypress

Shelob9:
@New0  has tested this manually with this filter https://gist.github.com/New0/43675dc394fef1510eba33668b7dcf5f

I did the same test, just modified the field ids in the code and it worked as expected.

Issue #2885, opened 2019-01-02
JavaScript Error `zh_cn is not available in the catalog`


Shelob9:
Reported in #2883

New0:
Closed via https://github.com/CalderaWP/Caldera-Forms/commit/ad312736b5197b947b05c34880ea4e4f59284980

PR #2886, opened 2019-01-03
Fixed: JavaScript Error `zh_cn is not available in the catalog` #2885

scrobbleme:
See issue #2885


New0:
Thank you @scrobbleme this was updated to the current state of develop branch via #3481 and issue closed via https://github.com/CalderaWP/Caldera-Forms/commit/ad312736b5197b947b05c34880ea4e4f59284980

Issue #2898, opened 2019-01-11
Merge 1.7.5-1.7.6 (master and develop-minor) to 1.8 (develop)


Shelob9:
The 1.7.5 and 1.7.6 releases need to be back ported into develop branch, so we don't lose those bug fixes. I will open a PR, as a start point.

Shelob9:
Close via 3b02be7 #2903

PR #2854, opened 2018-12-13
Refactor index.js to have testable functions

New0:

PR #2899, opened 2019-01-11
Merge master (1.7.5 and 1.7.6) to develop branch

Shelob9:
#2898


Shelob9:
The merge conflict here is not that bad. 

@New0 Does it make your life easier if we merge #2854 to develop, then deal with this merge, or would it be easier if you merge this into the branch you're working from?

My concern is that I don't want to lose these changes in 1.8, whatever solves that and makes your life easier, we do.

Issue #2904, opened 2019-01-14
Add 'disabled' attribute for Checkbox, Dropdown, and Autocomplete field types


mecachisenros:
## Do You Need Immediate Help?
No

## Is This A Feature Request?
Yes.

The Radio field type has the very useful `disabled` attribute which allows options to be disabled through Caldera Forms filter hooks like `caldera_forms_render_get_field`, the Checkbox, Dropdown, and Autocomplete field types lack the `disabled` attribute.

This is a request to add the `disabled` attribute to the mentioned field types.

### Example and use case
Both options for the Radio field **Event 1** are disabled (filtered through `caldera_forms_render_get_field` ) based on an external condition (in this case is an Event registration integrated with CiviCRM), the Option 2 option for the Event 2 fields (Checkbox, Autocomplete, and Dropdown) are disabled based on an external condition (in this case that option is Sold out).

#### Before
There's no possibility to achieve this with Checkbox, Dropdown, and Autocomplete field types.

#### After
It is possible to disable field options for the mentioned field types.

![disable_field_options](https://user-images.githubusercontent.com/3038096/51115818-371a1180-1801-11e9-9619-32e91a63f99a.gif)

Shelob9:
PR #2905 should resolve this.

New0:
Closed via 
 [`46141dc`](/CalderaWP/Caldera-Forms/commit/46141dc3329ad755f0583de968bf87878eff1b16)

PR #2905, opened 2019-01-14
#2904 add 'disabled' attribute for Checkbox, Dropdown, and Autocomplete field types

mecachisenros:
This PR adds the `disabled` attribute for the the Checkbox, Dropdown, and Autocomplete field types, as per my comments on #2904


Shelob9:
@mecachisenros Thanks for this! Looks good to me. We have 2 big PRs for 1.8.0 that are about to land and we can clean up this merge - don't worry about the conflicts until then.

@New0 I'm concerned about sneaking for things on to 1.8.0 because we need to ship, but this one is really simple, as long as it doesn't cause ghost inspector to fail, we should do it in 1.8.0. Let's discuss this tomorrow.

mecachisenros:
Many thanks @Shelob9!

I have one more suggestion/feature request regarding this, which I would like to hear your thoughts/opinion.

When disabling a Radio field’s option the default value (`checked` attribute) is [hard coded](
https://github.com/CalderaWP/Caldera-Forms/blob/master/fields/radio/field.php#L33) to `none/empty`, disabled fields/options are never submitted in a form so for me it doesn’t make sense to hard code it.

This can cause issues when viewing/editing an entry if you have conditionals based on an option and this option’s default has been reset.

I’m happy to add a second commit to this PR or open a separate issue with it’s corresponding PR, but since it’s just removing the `$checked = false` I thought it’s worth asking here.

Shelob9:
>I have one more suggestion/feature request regarding this, which I would like to hear your thoughts/opinion.
Can you please open a new ticket and a new PR for that? One change = one issue, one PR is the ideal.

Also:
We talked about this today - We will merge #2854 and then update this branch, if tests pass, then we'll merge in 1.8.0. If not, we will move it to 1.8.1 milestone and fix it once 1.8.0 is out. Need to ship soon!

mecachisenros:
Many thanks @Shelob9 PR #2914 and issue #2913 created.
I'll give it a go and also run the unit tests myself.

Shelob9:
@New0 Please review now that everything is merged, if it works, merge, if not, we'll re-address in 1.8.1

Shelob9:
Accidental close.

New0:
First test on a threee options checkboxes field breaks \:
[debug.log](https://github.com/CalderaWP/Caldera-Forms/files/2792155/debug.log)

mecachisenros:
@New0 my bad, I've amended it, I made the assumption `$disabled` was defined :-1: 
 
cb229bc should fix it, I've tested locally and ran `composer wp:tests` as well.

New0:
Update for @mecachisenros , sorry it seems like my commit was pushed (I thought it had failed so I requested the changes )

New0:
@Shelob9 this works for me if you want to include it to 1.8.0 again, you can see the Gist example of how to use it at https://gist.github.com/New0/43675dc394fef1510eba33668b7dcf5f

Shelob9:
@new0 I can't think of any reason not to merge for 1.8.0 besides I'm nervous and it isn't related to our 1.8.0 priorities. Checkbox fields, including getting better HTML markup in there finally, could literally be the1.8.1 focus.

If Ghost Inspector and Cypress tests pass and you want to merge for 1.8.0, go ahead, I have no rational reason not to, and want to make Kevin and friend's lives easier.

Issue #2913, opened 2019-01-16
Radio type field default value (i.e. checked) gets overridden when disabling its options  


mecachisenros:
## Do You Need Immediate Help?
No

## Is This A Feature Request?
Yes.

When disabling (setting 'disabled' attribute value to `true`) on a Radio field's options (via `caldera_forms_render_get_field` or other filters), its default value (`checked` attribute) is forced to `false`. This can cause various issues when viewing or editing an entry, or when providing a default value, and other fields have conditionals based on the Radio field's value.

Disabled fields/options are [never submitted](https://www.w3.org/TR/html401/interact/forms.html#disabled) in a form so for me it doesn’t make sense to force its value.

My suggestion is to remove the forced `checked` value.

New0:
Closed via 
[`ff2a14d`](/CalderaWP/Caldera-Forms/commit/ff2a14d2214239bc80782125d88181ae7a2b7504)

PR #2914, opened 2019-01-16
#2913 Remove forced 'checked' attribute value for Radio field types

mecachisenros:
As per my comments in #2913.


New0:
@Shelob9 The PR removes a line without alternatives, I ran tests on this branch and they all passed, do you have an objection before I merge that PR ?

Shelob9:
@New0 No objections

Issue #2912, opened 2019-01-16
CF2 File filed doesn't get unhidden by conditional logic


New0:
Found during ghost inspector test 5c3f47184f812d73c74d34f5 created for https://github.com/CalderaWP/Caldera-Forms/pull/2910#issue-244874506

The file field is hidden by a condition set from a checkbox field, after the checkbox is unchecked the file field is still not visible and doesn't get unhidden. ( I also tested with a switch field and different conditions settings like hide if element is / is not - show if element is / is not )

Note for today's context \: branch 2018-2 is currently merged to develop - a revert branch exists but was not merged yet - this means we can currently test from develop branch @Shelob9 should we let 2818-2 merged to develop and crate to branch to fix that issue from develop or create the branch to fix that issue from 2818-2 ?

New0:
First debug findings \:

- The template var l290 of conditionals.js only returns the wrapper div of the field when a condition is applied on a cf2 file field instead of the whole field html
( => The template var retrieves the html holded in script id '#conditional-' + field + '-tmpl' )

In my case this is stored \:
```
<script type="text/html" id="conditional-fld_6367931_1-tmpl">
<div id="cf2-fld_6367931_1" class="cf2-field-wrapper" data-field-id="fld_6367931_1" data-field="fld_6367931"></div>
</script>
```
When it should have stored something like\:
```
<script type="text/html" id="conditional-fld_6367931_1-tmpl">
<div id="cf2-fld_6367931_1" class="cf2-field-wrapper" data-field-id="fld_6367931_1" data-field="fld_6367931"><div class="form-group cf2-field-group"><label class="control-label" for="fld_6367931_1" id="fld_6367931_1Label">CF2 File field</label><div class="cf2-dropzone" data-field="fld_6367931"><div class="form-control cf2-field cf2-file" style="position: relative;" aria-disabled="false"><button type="button" class="btn btn-block" aria-controls="fld_6367931_1, cf2-list-files" aria-expanded="false">Drop files or click to select files to Upload</button><input type="file" accept="" id="fld_6367931_1" style="position: absolute; top: 0px; right: 0px; bottom: 0px; left: 0px; opacity: 0.00001; pointer-events: none;" autocomplete="off"></div></div></div></div>
</script>
```

That means that when the action gets to 'show', it returns the template, but in the case of a cf2 file field, it is incomplete.

Shelob9:
[...]

PR #2921, opened 2019-01-21
Bind field value change properly so that file field becomes unhidden

New0:
I opened this PR to keep a log of a fix for #2912 ( which already has PR #2916 ). 
Still can't figure out why minified assets doesn't give same result.
@Shelob9 It still only fix the issue with assets not minified


Shelob9:
@New0  75b2f50 "fixes" this. My "solution" was to copy the unmified version into https://www.minifier.org/ and then paste the result of that into /assets/js/build/conditionals.min.js 

Conclusion: Blame Grunt minify.

Issue #2922, opened 2019-01-22
Grunt concatenation and minification seems incomplete


New0:
PRs #2916 and #2921 show that Grunt process breaks CF2 conditional result.

New0:
Concatenating ajax-core and conditionals into caldera-forms-front.js fixes the issue.

This means when $should_minify os True we enqueue only caldera-forms-front.min.js instead of the three files ( caldera-forms-front.min.js, ajax-core.min.js and conditionals.min.js ).

Shelob9:
Close via 23dc378   #2923

PR #2923, opened 2019-01-22
concatenating ajax-core and conditionals into form front

New0:
#2922 

(This make PR #2921 work with normal Grunt process)


New0:
https://github.com/CalderaWP/Caldera-Forms/issues/2922#issuecomment-456343060

Issue #1765, opened 2017-07-21
If there is a non-required email field on page 2+ left empty, it silently fails validation


Shelob9:

Shelob9:
Probably close via f4f507e

PR #2928, opened 2019-01-22
Do not validate empty fields on page advance

### LINKED ISSUE #2911: When advancing pages via breadcrumbs - empty fields are validated, preventing advancing
**no**

Latest and latest 7.2, tried through Local by Flywheel too.

**Yes.**

Can't click to next page when clicking on the **breadcrumbs** link, does work when clicking on a 'next page' button.

Shelob9:
Fixes #2911

This removed the force option when violating. According to the docs: ` If force is given, it force validates even non-required fields `

There is a reason I added force:true there, I don't know if it is still valid, but I worry this is going to cause problems -- lets check the fancy fields (star/slider/files/etc.) do not break conditional hide/unhide and rebind after page advance.


Shelob9:
https://github.com/CalderaWP/Caldera-Forms/commit/779754ffc2a26c6fd578a4e81cb93427e3df74b4#diff-376e984735c05c485ba5e75257c9c524L141

https://github.com/CalderaWP/Caldera-Forms/commit/e1db1628283431b7c793db862c769247d215cdf6#diff-376e984735c05c485ba5e75257c9c524L152

#2021

#1765

Shelob9:
Thsi comment is most relevant - https://github.com/CalderaWP/Caldera-Forms/commit/f4f507e4b2f07a0da626e6326f9defe37307e241#r23233392

I think I switched to this method because there was a bug related to empty email fields in parsely, but we've updated parsely, so my concern about this PR is _probably_ invalid.

swapnilmanjrekar:
@Shelob9 I am not able to reproduce this issue as I successfully able navigate through breadcrumbs and also tested for the following scenarios which were as per my expectations:

-   Empty, non-required fields do not validate.
-   Empty, required fields do validate.
-   Non-empty fields do validate.
- Validation errors blocks submission.

WordPress Version: 5.0.3PHP Version: 7.1.26MySQL Version: 5.7.25Caldera Forms Version: 1.8.0-beta.1WP_DEBUG: 1

Issue #2895, opened 2019-01-11
CSV exported from entry viewer has JSON encoded options for checkbox fields


evolluti0n:
## Do You Need Immediate Help?
Ticket opened:
https://secure.helpscout.net/conversation/747675939/17690?folderId=2372292

## What Version Of Caldera Forms, WordPress and PHP Are You Using?
WordPress Version 5.0.2-de_DE
PHP Version 7.3
Caldera Forms Version 1.7.5.1

## Does Your Issue Persist When You Disable All Other Plugins and Switch To The Default Theme?
Yes

## What Is The Unexpected Behaviour?
I have a checkbox type field in a form. The field is exported as a coma separated list which is then repreated with {"opt ....}. See this example:

Winterball am 26.12.2019, 1 .Skitag fÃ¼r Frauen am 5.01.2019, Curling fÃ¼r Erwachsene am 12.01.2019, 1. Skitag fÃ¼r MÃ¤nner am 19.01.2019, 1. Gleitschirm-Tandemflug-Tag am 26.01.2018, Renntraining fÃ¼r Erwachsene am 2.02.2019, Clubabend am 9.02.2019, {"opt2017242":"Winterball am 26.12.2019","opt1302036":"1 .Skitag f\u00fcr Frauen am 5.01.2019","opt1156424":"Curling f\u00fcr Erwachsene am 12.01.2019","opt2001886":"1. Skitag f\u00fcr M\u00e4nner am 19.01.2019","opt1430941":"1. Gleitschirm-Tandemflug-Tag am 26.01.2018","opt1087773":"Renntraining f\u00fcr Erwachsene am 2.02.2019","opt1102332":"Clubabend am 9.02.2019"}

Is this a bug or is a normal behavior?

Shelob9:
@New0 I am adding this to 1.8.0 as I know what's wrong and its important to get fixed.

Shelob9:
Related(ish) #1032 #2078 #374 #173

Shelob9:
Should be resolved via 
6fb6bc8 when #2919 is merged

Shelob9:
I updated the title to avoid the confusion noted here: https://github.com/CalderaWP/Caldera-Forms/issues/2895

New0:
Closed via [`27e20f2`](/CalderaWP/Caldera-Forms/commit/27e20f20399d31d7dfc240478e24e5d432691d11)

PR #2930, opened 2019-01-24
Fix checkbox field values appearance in CSV file

New0:
Fix for #2895

### DIFF
--- a/classes/magic/doer.php
+++ b/classes/magic/doer.php
@@ -550,6 +550,9 @@ class Caldera_Forms_Magic_Doer {
 				$value = implode( ', ', (array) $_value );
 			}
 
+		} else if( is_array( $value )  ){
+
+			$value = implode( ', ', $value );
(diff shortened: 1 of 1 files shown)


Shelob9:
@New0 Your solution is smarter. My solution addressed the values coming into CSV exporting wrong - https://github.com/CalderaWP/Caldera-Forms/commit/6fb6bc8c10962cbebeadfa613bd05568635d5d88#diff-9f9968437cfc01236a148f223bb2d071R1546

Your solution prevents them from being wrong in the first place.

Shelob9:
I checked two boxes in my test form, in the entry viewer I see the first option, in the csv i get nothing for that field. I suspect my test setup is bad. I added a form to the tests spreadsheat called `2895 Checkboxes` that's what I'm testing with.

screen shot 2019-01-24 at 1 17 44 pm

PR #2933, opened 2019-01-25
Improve file field max size errors and validation

### LINKED ISSUE #2932: Errors for too large of a file are insufficent
File size limitation is added in #2823 to fix #1399 

A file that is too large should be rejected in the client when its added.

It is already rejected server side resulting in an error response here -- https://github.com/CalderaWP/Caldera-Forms/blob/develop/cf2/RestApi/File/CreateFile.php#L126-L134

Let's make sure that if a file is added to a field that is too large -- an error is shown on the field and the offending item in the list is highlighted in red and the form can't be submitted until its removed. It's tempting to say it should just not add that file, but then a user might add a file, see it didn't get added, add it again, see it didn't get added and get annoyed. They should see what is wrong clearly.

Let's make sure if file upload fails, it opens that notice and stops form submission -- second part is already covered. 
https://github.com/CalderaWP/Caldera-Forms/blob/master/assets/js/ajax-core.js#L82

New0:
Fix #2932


Shelob9:
I'm running tests on develop branch right now. I see errors like

 ```
calderawp\calderaforms\Tests\Integration\Caldera_Forms_FilesTest::testIfFileIsTooLarge
Illegal string offset 'size'

/app/classes/files.php:327
/app/tests/Integration/Caldera_Forms_FilesTest.php:217
```
Let's make sure this PR addresses that.

Shelob9:
The error message needs to be meaningful. A message like `sprintf ( 'File %s is too large. Maxium size is %s, $fileName, $fileSizeInMb);` 

If a field has an error, it should be invalid -- therefore you should not be able to submit form.

New0:
@Shelob9 I thought I could use setBlocking function from clients/render/utils.js to block submit button to start request, but that's not the case, do you confirm it's normal and I should use another way ?

Shelob9:
@New0 If `setBlocking()` does not block submission, that is a bug.

New0:
> 
> 
> @New0 If `setBlocking()` does not block submission, that is a bug.

Just to be clear, as long as there are elements listed in cf2.fieldsBlocking array , the Submit button shouldn't start request, right ? This would be normal behavior.

Shelob9:
You are correct

New0:
First Ghost inspector test https://app.ghostinspector.com/tests/5c5859f9a07feb5157ea6a7c

PR #2935, opened 2019-01-25
Fix PHP 7.3 warning

### LINKED ISSUE #2892: Throws warning line 122 in /classes/magic/summary.php
/classes/magic/summary.php line 122 throws a warning due to changes in PHP 7.3.

https://make.wordpress.org/core/2018/10/15/wordpress-and-php-7-3/

rebeccahum:
Resolves #2892.


Shelob9:
@rebeccahum Thank you for submitting this pull request. This looks good, but we need to test it before we can merge, I'm a little concerned that using continue there may have unintended consequences.

I have placed this on the 1.8.1 milestone. We are not adding any new issues 1.8.0 as we want to get that shipped and can't if we have to review more PRs. We expect 1.8.1 to follow 1.8.0 pretty quickly as there are a few open PRs like this that we will want to ship right away.

PR #2936, opened 2019-01-28
#2894 Add HTTP referrer magic tag

### LINKED ISSUE #2894: HTTP Referrer Ability
* If so, please open a support ticket at https://CalderaForms.com/support

## Is This A Feature Request?
* If so, that's fine, please describe your desired featured and delete the rest of this template.

Please create an HTTP Referrer ability. Example: _Customer clicks on a link in a Facebook post that takes them to a page that has a Caldera Form, they fill it out, then in the email the website admin receives is the url the customer lead came from_. This provides a ton of value to the clients and helps Agency's with Caldera Forms stand out. This is a pretty standard feature for wp contact forms (contactform7, gravity form) so I think it would be pretty easy feature to quickly add.

tmdesigned:
Adding referrer_url magic tag to capture HTTP referrer from either WordPress tracking (if present) or $_REQUEST (if present). 

Utilizes WP built-in function wp_get_raw_referer() rather than the non-raw version because there is no need to filter against a white-list as we are not redirecting to it.

Both referrer (proper spelling) and referer (typo present in WP) are included for parsing tags. Only referrer is included for the autocomplete suggestion. 

Fixes #2894


Shelob9:
thanks for the pull request. FYI we are finishing 1.8.0 now and are not going to add additional features to that milestone. We will review this for inclusion in 1.8.1 or 1.9.0.

New0:
Thank you for the PR @tmdesigned , I changed spelling for dropdown option use as WordPress function spelling is "referer" 
Also I would use "wp_get_referer" as wp_get_raw_referer" is not validated before saving to database.
@Shelob9 what do you think for both points ?

tmdesigned:
Thanks @New0. It looks like the difference between "wp_get_raw_referer" and "wp_get_referer" is that the second runs the first through an additional function, "wp_validate_redirect", which is designed to allow URLs to be compared against a whitelist prior to redirection. But in this case we aren't redirecting, and I don't think wp_get_referer is sanitizing for the db either (I may be missing it).

tmdesigned:
I do think it needs to be sanitized prior, as both wp_get_referer and wp_get_raw_referer are just getting the value from $_REQUEST or $_SERVER. I have added a commit with esc_url_raw

New0:
Thanks again @tmdesigned and sorry for the delay.

Issue #2938, opened 2019-01-28
Add a color option for cf2 File Field button


New0:
What's missing now for this feature request is a hover state management.

This can be done by adding a package that allows css in js to use ':hover' ( Like https://www.npmjs.com/package/react-hover-observer or https://www.npmjs.com/package/radium ) or add a state management in clients/render/components/CalderaFormsRender.js

Shelob9:
@New0  Should this be on 1.8.0 or 1.8.1 ?

PR #2939, opened 2019-01-28
Add contrast to the button on hover

### LINKED ISSUE #2832: Hover text of file field button is illegible.
screen shot 2018-11-27 at 10 28 02 am
No way that contrast passes a WCAG scan.

New0:
Fixes #2832


New0:
> 
> 
> @New0 Have you tested the screen contrast with 2016 and 2017 using a WCAG color contrast tester? We should make sure that regular and hover state have the right contrast.

Yes it passes accessibility tests, although it defines fixed color ( the dark grey used in previous File fields ), and I think we could add this color as an option, let's consider this for future release => https://github.com/CalderaWP/Caldera-Forms/issues/2938

Issue #2950, opened 2019-01-29
The file list under advanced file field v2 is VERY different from v1


Shelob9:
I have reviewed the file fields more throughly and I think that there are too many difference in v1 vs v2 fields and the differences are likely to cause confusion, so I am considering this issue to be a blocker.

This is a v1 and a v2 file field next to each other:
![screen shot 2019-01-29 at 2 53 18 pm](https://user-images.githubusercontent.com/1994311/51936211-a84aee80-23d5-11e9-9888-de67b887713a.png)

I think the one on the right looks better. The markup we have now is more semantic, but it does not look as good. Let's keep the same HTML markup if possible.

- [ ] Make the list of files in the field use the same typography and font/icon sizes as v1. 

 This screenshot I enabled image preview for the v2 field on the right, after adding a file to the v1 field on the right, I used the inspector to replace the name of the file with the image preview. Looks better. 

![screen shot 2019-01-29 at 2 57 24 pm](https://user-images.githubusercontent.com/1994311/51936621-9ddd2480-23d6-11e9-9b9e-2a5c7a6d5ac3.png)

- [ ] When using image preview, replace image name with image preview.

Also, v1 file fields list file size in kB, MB or GB. v2 fields are using bytes, which is confusing.

- [ ] File size should be listed in kB, MB or GB.

When you submit a form with a v1 file in it, you get a little progress bar that indicates that something is happening. With v2 there is no visual indication. This is likely to lead to multiple submissions as people will just click submit until it looks like something happened.

- [ ] File fields should show a progress indicator when uploading.

New0:
Closed via https://github.com/CalderaWP/Caldera-Forms/commit/a27a239bb1854e80711705f624460a1df265c73c

Issue #2947, opened 2019-01-29
Frontend CSS for clients should be loaded from clients 


New0:
The Grunt task concatenates the files from clients/render into assets/build/css/caldera-forms-front.min.css 
As we want to avoid assets folder, we should load them from clients/render folder.

Shelob9:
`Caldera_Forms_Render_Assets::enqueue_style( 'render');` should be all you need.

Shelob9:
close via b3c27fb #2952

PR #2952, opened 2019-01-30
Make CF2 File field look like Advanced FIle Field 1

New0:
This PR works on points of #2950, these are the main 4 \:

- [x] Make the list of files in the field use the same typography and font/icon sizes as v1.

- [x] When using image preview, replace image name with image preview.

- [x] File size should be listed in kB, MB or GB.

@Shelob9 I agree with all the points, just need to add that the progress bar will require more time


Shelob9:
@New0 BTW This is the function we're currently using for file size https://github.com/CalderaWP/Caldera-Forms/blob/master/fields/advanced_file/uploader.js#L2

New0:
@Shelob9 I left the button to remove the files a bit bigger than v1, I think it too small to be accessible, could you review and let me know what's missing to check off "Make the list of files in the field use the same typography and font/icon sizes as v1." 

- >Just a reminder to test after flushing browser cache (ctrl + f5 on win), to be sure to load last css

Thanks for the function.

Shelob9:
On the left  is what I see when I 
* checkout this branch at 7f1ade3 
* run `npm run build:webpack`
* Flush browser cache
* add a file to field.

screen shot 2019-01-30 at 9 58 32 am

On the right is a v1 field, same theme. http://dev-caldera-testing.pantheonsite.io/2021-adv-file-created1-5-7-1-dev1-5-8/

I think that the one on the left has the following differences that we should recreate in the new field:
* smaller text
* a space between the X and the name of the file
* a lighter underline
* the file size displayed in menainful units
* the file type and size under the name
* The X looks like something I should click on.

Shelob9:
>I think it too small to be accessible
I checked http://dev-caldera-testing.pantheonsite.io/2021-adv-file-created1-5-7-1-dev1-5-8/ with https://achecker.ca/checker/index.php and it did not raise issues there. Does your testing raise any errors on that?

New0:
@Shelob9 we are facing issue https://github.com/CalderaWP/Caldera-Forms/issues/2947 

At the moment we need to run `npm run build` and after again `grunt` for that Stylesheet to be loaded on frontend. 

This is how I see it 
screenshot_2019-01-30 2944 dev 1

Shelob9:
Ok, if that's what it looks like, that's great. I liked the smaller X, but
if you think it is too small, then that's fine by me.

Shelob9:
> File fields should show a progress indicator when uploading.
Trigger spinner when upload starts -- this is better than v1, where it starts after upload, which could create user confusion as they might not see progress bar.

Shelob9:
@New0 Please resolve conflicts and merge.

Issue #2962, opened 2019-02-04
Overzealous magic tag processing - strings with 2 or more % get filtered


tmdesigned:
## Do You Need Immediate Help?
No

## Is This A Feature Request?
No

## What Version Of Caldera Forms, WordPress and PHP Are You Using?
WordPress Version: 5.0.3PHP Version: 7.3.0MySQL Version: 5.5.5Caldera Forms Version: 1.7.6WP_DEBUG:

## Does Your Issue Persist When You Disable All Other Plugins and Switch To The Default Theme?
Yes

## What Is The Unexpected Behaviour?
Any value with two or more % signs in the text gets treated as having a field-based Magic Tag, even when there are no matches, and the values between the % pairs get removed.

This is especially problematic with capturing URLs as they often contain many url-encoded values with %, i.e. ?query=my%20search%20phrase becomes ?query=my20phrase

## Expected Behavior
If a value matches a magic tag source, replace it. If not, leave it alone.

**NOTE** bracket magic tags already exhibit the desired behavior. Line 444 contains a check to see if the filtered_value is any different from the initial value

```
				if ( $filter_value != $magics[ 1 ][ $magic_key ] ) {
					$value = str_replace( $magics[ 0 ][ $magic_key ], $filter_value, $value );
				}
```

tmdesigned:
I've submitted PR #2963 to address this. Like I mentioned there I realize there is a significant impact if users are depending on the existing behavior.

Also, per Issue #2960 the fields will still get stripped in the email as there is similar processing that is done outside of the context of the magic tag class.

PR #2963, opened 2019-02-04
#2962 - Do not replace magic field tags if no match found

tmdesigned:
Bracket tags already do this. This proposed change prevents data loss/corruption when 2 or more % are found in the value being processed by the magic tag doer. 

Even though I think this should be the default behavior, I recognize this has a moderate impact and people may be depending on the current behavior. May need to be more of a discussion.


Shelob9:
@tmdesigned Thanks for the PR. We have frozen 1.8.0 for new features. We will consider this for 1.8.1 or 1.9.0 once we are done with 1.8.0

>Even though I think this should be the default behavior, I recognize this has a moderate impact and people may be depending on the current behavior. May need to be more of a discussion.

Can you please explain in the ticket what you mean by "moderate impact"?

Shelob9:
Sorry, accidental close.

tmdesigned:
I'm not *that* familiar with the Caldera Forms code base, but in researching this issue there were a **lot** of places that called this function. It just makes me hesitant because I haven't looked at each and every one of them. It "should" make sense to always leave the initial value if it can't find a valid replacement. But you would know better than I. 

I can also see if someone had discovered this behavior in the past they may have started counting on it. Perhaps they copy the same form across dozens of sites and put in a bunch of magic tags, knowing that the ones that aren't there will be filtered out and not show up. 🤷‍♂️

alexylem:
Super annoying as it breaks all links (url parameters) in my CF notification emails...
Any plan to have the proposed fix merged?
Didn't find another way around this, or is it possible to completely turn-off variable replacement ? Or choose to use another char than % ?

Issue #2964, opened 2019-02-04
Unhide cf2 field creates error


New0:
Ghost inspector test https://app.ghostinspector.com/tests/5c3882b70b45715fbf380071 is failing after showing `"Warning: Failed prop type: Invalid prop `field.fieldValue` supplied to `CalderaFormsFieldRender`.\n in CalderaFormsFieldRender (created by CalderaFormsRender)\n in CalderaFormsRender"`

Shelob9:
Close via f5c5a8f #2965

PR #2965, opened 2019-02-04
Unhide a cf2 field with correct field value set

New0:
Fix for #2964 It seems that the getFieldValues was used to set a field value during hide/unhide process but it returns all fields values and creates an error when unhiding a field.

In this PR I set the field Value from the field state values


New0:
@Shelob9 I renamed the function and used your solution that makes more sense

Issue #2948, opened 2019-01-29
If there are two adv file v2.0 fields, form submission generates JavaScript error


Shelob9:
To recreate:

Create a form with two advanced file fields. Submit the form with files in both forms.

The form submits. Data is saved correctly(ish) -- I'll make a separate issue -- There are no PHP errors logged. The Javascript console generates error `Cannot read property 'data' of undefined` from render client.

This screenshot shows the issue using the form I uploaded here: https://github.com/CalderaWP/Caldera-Forms/issues/2944#issuecomment-458675488
![screen shot 2019-01-29 at 2 28 57 pm](https://user-images.githubusercontent.com/1994311/51934737-4a68d780-23d2-11e9-8767-af07ea5e85f7.png)

#2944

Shelob9:
We should not work on this until #2933 is merged.

Shelob9:
This is unblocked as  #2933 is merged.

Shelob9:
Notes: 
1) if i leave on file field empty, it submits as expected.
2) If i put files in both fields, neither field triggers a request to upload, and as a result, but it does mark them as pending, so ajax-core blocks submission.
3) We can to remove ajax-core.js as soon as we remove advanced file fields v1. In hindsight we should have done that first.

I checked out from develop branch at commit a3cd7f6ef69534db8b2ddca1107bb8bf065f4ade

WordPress Version: 5.0.3PHP Version: 7.1.25MySQL Version: 5.7.24Caldera Forms Version: 1.8.0-alpha.1WP_DEBUG: 1

Shelob9:
I can't get consistent results watching these breakpoints in `webpack:///./clients/render/fileUploads.js`
![screen shot 2019-02-06 at 5 10 44 pm](https://user-images.githubusercontent.com/1994311/52377236-496b2200-2a32-11e9-9d37-b772984a939a.png)

With a form with one file field, debugger hits on L194, then L91, then 103, then 106 or 110. That's how it should work.

With a form with two file fields, sometimes it does upload the first field if the second field is empty, sometimes it does not -- IE L194 hits then 103, then nothing.

LOGIC CONCLUSION -> We are not controlling async execution properly. We should `await` the result of file hashing.

Shelob9:
Close via aa0cce8

PR #2971, opened 2019-02-07
Prevent ajax-core to trigger errors when multiple CF2 file fields are submitting

New0:
Check if all fields values have been processed before calling $form.submit, in other words, only call $form.submit once

Work for #2948


New0:
@Shelob9  I had not commit the build files, could you test again please ?

The form shared only have one file field, but I tested again with two files field successfully.

Issue #2764, opened 2018-10-16
Overview: Refactor testing and CI


Shelob9:
Types of tests:
1. Integration tests WordPress "Unit" Tests in CF with phpunit 
- Currently runs in Travis
2. JavaScript unit tests in CF
- Works, need to automate
3. Ghost Inspector Testing
- Manual, will stay that way.
4. Cypress e2e tests
- Works in https://github.com/CalderaWP/calderawp-wordpress-plugin
5. Isolated unit testing
- Works in https://github.com/CalderaWP/calderawp-wordpress-plugin

TODO:
- [ ] Copy the docker setup from wordpress plugin
- [ ] Copy the installer scripts (bash/npm/composer)
- [ ] Install cypress in Caldera Forms
- [ ] Copy cypress 2e2 test into Caldera Forms.
- [ ] Setup isolated unit tests in CF
- [ ] Make existing integration tests work with Docker and the way Travis currently does it.
- [ ] Run JavaScript unit tests in Travis
- [ ] Run isolated unit tests in Travis
- [ ] In Travis, run JavaScript unit tests, then phpunit unit tests, then WordPress integration tests, then build environment and run e2e tests (possibly only on some versions for last step)
- [ ] Document it all

PR #2974, opened 2019-02-08
Make sure an event will cause effect on the component of current form and not same component in another form in the page

### LINKED ISSUE #2972: When two same forms are inserted in a page, the cf2 fields rendered have duplicated behavior
In the case that we insert twice the same form in a page, same CF2 field are rendered twice and their state are mixed.
The value and the error messages will be shared.

New0:
Check that the form id in event currentTarget matches the component form id

This prevents messages or submissions accross same form inserted twice in same page

This fixes #2972


Shelob9:
@New0 Please fix conflicts and merge if tests pass after rebuilding.

Shelob9:
Related to Travis issue, locally I get error `fatal: No url found for submodule path 'vendor/calderawp/caldera-forms-query' in .gitmodules` Composer checks it out with the .git dir, but there is no submodules folder.

On Travis we use `composer update --prefer-dist` which prevents it from checking out via .git, but beacuse git thinks its a submodules, I don't really know, but it shouldn't be a submodule so I used `git rm -rf` to remove it in f01192d

Shelob9:
My previous comment, not the issue. Glad I fixed it, but not the problem. Problem is we are using phpunit 5, beacuse we supported PHP 5.2 and that doesn't work well with PHP 7.2. We should probably switch to phpunit 7. For now, let's just not test PHP 7.2, so we can be done (768008f)

Anyway @New0 You should be good to finish this. #2982 may have similar problems. You may want to make the travis changes there to.

New0:
Thank you @Shelob9 let's address Travis for php 7.2 in issue #2764

PR #2982, opened 2019-02-13
Validate cf2 file fields on page navigation

### LINKED ISSUE #2981: cf2 file field does not block advancing page if invalid
I have a form with a file field v2 on page 1 along with a next page button, and a submit button on page 2. The file field is required, the file field only allows png files. I am able to advance the form to page 2 while the file field in page 1 is invalid by:
1. Clicking the next page button
2. Clicking the breadcrumbs link to page 2.

I am making the file invalid by:
1. Adding an svg file to the field -- only png is allowed.
2. Clicking the submit button on page 2, which does trigger validation.

There is an event we can hook into, I believe, or we can add one.

Shelob9:
Fixes #2981

Includes PR #2974 which should get merged first, so I'm making this easier.

In this PR, I moved the data-field attribute for the CF2 file field from wrapper to the field element. Then I added the component ref to the window.cf2 object so I could use it to validate cf2 fields on page nav.


Shelob9:
Once PR #2974 is merged you should finish this. Finishing it will require accepting snapshot changes that I want you to double check first.

Shelob9:
TL;DR `git rm -rf vendor`, switch caldera-forms-query to 0.3.2 tag, `composer update --prefer-dist`

Hopefully this is the end of this one.

BTW 3f1ddd5 92bcb8b and 37a7782a61b537fc442425348984a413313326df got pushed to a already merged branch, so re-did in correct branch with 40568ca  31b067e and 501e4ed

Shelob9:
https://getcomposer.org/doc/faqs/should-i-commit-the-dependencies-in-my-vendor-directory.md

PR #3001, opened 2019-02-20
get cf id correctly during upload of adv file v1 fields 

### LINKED ISSUE #3000: Adv File Field v1 do not upload
If a form has the version 1 advanced file field, the file upload does not route correctly and form submission does not complete.

class `Caldera_Forms` L3571, it returns for reasons I don't get. I think that is the issue.

screen shot 2019-02-19 at 8 46 56 pm

Shelob9:
adv fiel field v1 forms were not uploading correctly. This PR improves how we find CF id during submission and it also adds success => true to the result.

Fixes #3000

### DIFF
--- a/classes/core.php
+++ b/classes/core.php
@@ -3561,23 +3561,25 @@ class Caldera_Forms
 	{
 
 		global $wp_query;
-		if (isset($wp_query->query_vars['cf_api']) ||  isset( $_GET, $_GET[ 'cf-api'])) {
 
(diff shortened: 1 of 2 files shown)


Shelob9:
Also, because the new function `caldera_forms_send_json` does not use `wp_send_json_sucess` I added a line to always return sucess => true.

BTW @New0 I wrote this on a plane while tired and did not test great. Please review with suspicion.

Issue #2653, opened 2018-06-29
'This file type is not allowed. Please try another.' is untranslatable. 


cicichirinos:
see https://wordpress.org/support/topic/missing-sentence-in-caldera-forms-pot/

New0:
Fixed via 
[`f9e52b0`](/CalderaWP/Caldera-Forms/commit/f9e52b0f96abdbae744269bf8b4c0c35ebe6ac42)

PR #3015, opened 2019-02-25
Update .pot file for translation reference

New0:
Fix for #2653 

The string was already translatable but pot file for translations reference was not updated


freakpants:
Please merge this issue. We cannot translate error strings for the better phone number field, because those are not yet included in the template file. There are also about 600 strings in the german translation that have been translated but are still in the status "waiting", some of them since last march.

https://translate.wordpress.org/locale/de/default/wp-plugins/caldera-forms

Issue #2929, opened 2019-01-23
In block, add a link to form editor for that form


Shelob9:
@cicichirinos would like to be able to get from the Caldera Forms block to the form editor more directly.

Placeing a link to the currently selected from -- using the `<ExternalLink>` component from `@wordpress/components` under the form selector in the InspectorControls will solve this.

New0:
Closed via https://github.com/CalderaWP/Caldera-Forms/commit/2555e23a6ee9d7a7e5c9d94270fce8fa582eb789

PR #3018, opened 2019-02-26
 Add a link from Gutenberg block to Form editor page 

New0:
This is for #2929


New0:
@Shelob9 I didn't use the <ExternalLink> component as it seems to be for links out of the current installation/website and it uses target="_blank" to open another tab, let me know if that's what you wanted.

Issue #3017, opened 2019-02-26
Dependency is not PHP7 compatible


Shelob9:
```
PHP error in: caldera-forms/trunk/vendor/symfony/contracts/Cache/CacheTrait.php:
				
				Parse error: syntax error, unexpected '?', expecting variable (T_VARIABLE) in caldera-forms/trunk/vendor/symfony/contracts/Cache/CacheTrait.php on line 40
				Errors parsing caldera-forms/trunk/vendor/symfony/contracts/Cache/CacheTrait.php
```

Shelob9:
In `vendor/symfony/contracts/Cache/CacheTrait.php`    on L42, I converted it to a regular ternary  `if (0 > $beta = $beta ? 1.0 : 0 ) {`

This file is not used, it got added as a dependency of Carbon, which is a dependency of wp-queue. So this PROBABLY does not mater.

PR #3019, opened 2019-02-26
PHP 5.6 compat in dependency

Shelob9:
This requires  "symfony/translation": "~3.0" to stay php 5.6 compat, previously we had 4.0, which is PHP 7.1+. I fixed this on WordPress.org in 1.8.0 while shipping. This PR will prevent same error when releasing 1.8.1

#3017


Shelob9:
@New0  I am merging this as it is a blocker to release of 1.8.1 and its already in master.

Issue #2978, opened 2019-02-11
Variable Pricing Form Fails Calculation after saving


PJP1111:
## Do You Need Immediate Help?
Hopefully soon

## Is This A Feature Request?
No

## What Version Of Caldera Forms, WordPress and PHP Are You Using?
Short Debug Information
WordPress Version: 5.0.3 PHP Version: 7.2.14 MySQL Version: 10.1.37 Caldera Forms Version: 1.7.6WP_DEBUG:

## Does Your Issue Persist When You Disable All Other Plugins and Switch To The Default Theme?
Yes

## What Is The Unexpected Behaviour?
Created a new form with the pricing template. When I preview it before doing anything else it works properly, but when I save the form (even without changing anything) it will not calculate. Instead it just adds all the values and ignores the form choices. Here is a short video showing what happens: http://www.powerjamproductions.com/calderabug1.avi

## What PHP Errors Have You Logged While Reproducing This Bug?
* See: https://calderaforms.com/2016/05/wordpress-debug-logging/
None

## What JavaScript Errors Have You Seen While Reproducing This Bug?
* See: http://webmasters.stackexchange.com/questions/8525/how-to-open-the-javascript-console-in-different-browsers
None

New0:
Hello @PJP1111 ,

Thank you for reporting this issue. 
It seems like the conditions of the form example failed, we are going to update them on next update.

In the meantime, I attached the form example as a zip file. After unzipping , you will have the json file that can be used to import a form into Caldera Forms.
[variable-price-example-export.zip](https://github.com/CalderaWP/Caldera-Forms/files/2910090/variable-price-example-export.zip)

New0:
Fixed via https://github.com/CalderaWP/Caldera-Forms/commit/134305700b3fbf7895211bae04ca5475f52a7587

PR #3020, opened 2019-02-27
Re-create conditions for the Variable pricing example form

New0:
Conditions for variable pricing example form are failing, I re-created and integrated them in php file that loads the example.

For #2978


Shelob9:
@New0 It's easiest to fix  #3023 in this PR and probably  best as it proves the fix.

PR #3038, opened 2019-03-01
Fix Datepicker styles

New0:
Removed fixed width for calendar columns, that way the first and last columns in the calendar shows correctly two figures numbers

Work for #3030

### DIFF
--- a/fields/date_picker/css/datepicker.css
+++ b/fields/date_picker/css/datepicker.css
@@ -101,7 +101,6 @@
 .cfdatepicker table tr td,
 .cfdatepicker table tr th {
   text-align: center;
-  /*width: 10px;*/
   height: 10px;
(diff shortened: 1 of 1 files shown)


fields/date_picker/css/datepicker.css
```diff
@@ -101,7 +101,7 @@
 .cfdatepicker table tr td,
 .cfdatepicker table tr th {
   text-align: center;
-  width: 10px;
+  /*width: 10px;*/
```
Shelob9: @New0 I'd rather we delete than comment out.

This is a nitpick -- IE it's fine, don't worry about it -- try and keep this in mind in the future.

New0:
Accidentally open PR against master replaced by #3054

PR #3045, opened 2019-03-04
Set correct value to html during init and events for Range slider field

### LINKED ISSUE #3043: Range Slider field displays Field Id Attr instead of value
Since update to 1.8.0 the Range slider field seems to display the Field Id Attr when it should display the Value, this does not have consequences on calculations

screenshot_2019-03-04 form - dev

New0:
After events, the html printed the field Id attr instead of value, this PR needs Ghost Inspector test before merging
This fixes #3043


assets/js/field-config.js
```diff
@@ -268,8 +268,12 @@
              };
 
              rangeSliders[field.id].init = init;
-             state.events().subscribe(field.id, function (value) {
-                 $('#' + field.id + '_value').html(value);
+             state.events().subscribe(field.id, function ( eventFieldIdArray, value ) {
```
Shelob9: @new0 Why `eventFieldIdArray` an array? 

If my memory is correct, sometimes are an array, and I never did anything with it because we didn't have test coverage. If this is the safest fix, and it works - great. I just want to make sure we're considering why its an  array and if it makes sense to fix that.

More importantly, does this cause problems in the callback? I notice you're not checking something like `if( ! field.id === eventFieldIdArray[0]){ return;}`... Maybe I'm over-protecting here.

New0: @Shelob9 we are not using the first parameter anymore, I renamed it eventFieldIdArray to be more explicit. 
eventFieldIdArray was previously named value and was set to fill the html span for value, that printed the fieldIdAttr when an event (like sliding) started.
Now I'm checking if value is not empty, if it is I define value with the field.default value.

Let me know if want me to investigate why it's an array anyway ?

Shelob9: That makes sense to me. I don't think correcting the array/string issue right now makes sense. Based on what you said, I think it's fine. i wanted to make sure you'd thaught this threw.

Issue #3083, opened 2019-03-12
Custom field used for calculation values always return the field's 'value'


mecachisenros:
## Do You Need Immediate Help?
No

## Is This A Feature Request?
Yes

When registering a custom field that implements html `option`s with their corresponding `calc_value`s via `caldera_forms_get_field_types` and using this field as a calculation value, the option's calculation value always returns the fields `value` (submitted value) rather that the `calc_value`.

This happens because `Caldera_Forms_Field_Calculation ::get_value()` [defaults](https://github.com/CalderaWP/Caldera-Forms/blob/d2f83e313aee8d0a2c9d2b99bd157684a4f44052/classes/field/calculation.php#L88) to the field's `value` (submitted value) if the field type is not a core field type ( ie. `checkbox`, `radio`, `dropdown`, etc.).

I propose to introduce filter for the calculation value, something like:
``` php
...
/**
 * Filter calculation field value.
 *
 * @since 1.8.x
 * 
 * @param float $number The calculation value
 * @param string|int The field entry value
 * @param array $field The field entry config
 * @param array $form The form config
 */
$number = apply_filters( 'caldera_forms_calculation_field_get_value', $number, $entry_value, $field, $form );

return floatval( $number );
...
```

## What Version Of Caldera Forms, WordPress and PHP Are You Using?
WordPress Version: 5.1
PHP Version: 7.2.14
MySQL Version: 5.7.25
Caldera Forms Version: 1.8.2

New0:
Hello @mecachisenros , thank you for your contribution. I agree with your scenario, the filter would solve this limitation.

@Shelob9 I also wonder if we shouldn't default to find_calc_value() and fallback to find_value() if empty as we are in Caldera_Forms_Field_Calculation Class, what do you think ?

Shelob9:
I agree with @New0 let's come back to this when we get 1.8.x stabler.

mecachisenros:
@New0 @Shelob9 thank you for considering this.

@new0 that's a very good point (but I think is the other way around), the default is `find_value()` but defaulting to `find_calc_value()` will allow to use `caldera_forms_get_option_calculation_value` filter from `Caldera_Forms_Field_Util::get_option_calculation_value()`.

New0:
@mecachisenros , the point was to avoid using a filter hook, you wouldn't need a filter hook if the default was find_calc_value()

New0:
[...]

PR #3084, opened 2019-03-12
#3083 add 'caldera_forms_calculation_field_get_value' filter

mecachisenros:
Proposal to add `caldera_forms_calculation_field_get_value` filter for filtering calculation field values as per description in #3083


kcristiano:
@New0 Can you let us know what we nee to do to get this merged in?   We've been running in production since 1.8.7 and would prefer not to have to maintain our own build

New0:
@Shelob9 could you review and also check how we can run CI tests on this branch?

Shelob9:
> @Shelob9 could you review and also check how we can run CI tests on this branch?
@New0 It looks good to me. I don't know how to run CI right for this. Let's merge it, run CI on develop, and if it breaks anything which I doubt it will, we can revert, but also, it's fine.

New0:
Thanks again @mecachisenros

PR #3087, opened 2019-03-12
create caldera_forms_print_translation_strings_in_footer filter 

Shelob9:
By default this does nothing. If the filter is set to return true, then the Parsely validtor strings will be printed in the footer.

`add_filter( 'caldera_forms_print_translation_strings_in_footer', '__return_true' );`

#2962


Shelob9:
Screen Shot 2019-03-12 at 4 03 37 PM

Shelob9:
Built ZIP file: [1.8.3-alpha.1.zip](https://github.com/CalderaWP/Caldera-Forms/files/2958627/1.8.3-alpha.1.zip)

Shelob9:
New theory, this approach is too complex.

The problem we are trying to solve is Parsely defaults to these strings https://github.com/CalderaWP/Caldera-Forms/blob/master/assets/js/i18n/en.js for validation errors, sometimes it switches out to other languages, so Weglot translates other languages. That makes no sense.

Proposal: If this filter is true, we pass those English strings to wp_localize_script -- Weglot's proposed solution. Else, it works the same way it always did. AND we can put a filter on those strings, which is a requested feature, I'll have to find the issue.

Shelob9:
With `e073931` the strings are now printed using wp_localize_script(). IF this works for Weglot -- with that filter set to return true. Then I need to make sure that they are actually the strings that get used by parsely. I think we need to replace `"<script> setTimeout(function(){window.Parsley.setLocale('$code'); }, 2000 );</script>";` with something like:

`window.Parsely.addMessages( $locale, CF_VALIDATOR_STRINGS ); window.Parsley.setLocale($locale); }`

Shelob9:
[1.8.3-alpha.1.zip](https://github.com/CalderaWP/Caldera-Forms/files/2962859/1.8.3-alpha.1.zip) New build including the last commit.

Shelob9:
@new0 your correction is correct. 

I want to make sure these strings are used by Parsely.

Shelob9:
@New0 I refactored a bit in 581174b as this wasn't working for me because my language is `en_US` and so it was returning early. I confirmed this worked by setting the filter to true, changing the required field message and submitting the form. 

In my opinion this should be merged now.

Issue #3091, opened 2019-03-14
Custom CSS Class not applied in Gutenberg block


tmdesigned:
## Do You Need Immediate Help?
No

## Is This A Feature Request?
No

## What Version Of Caldera Forms, WordPress and PHP Are You Using?
WordPress Version: 5.1PHP Version: 7.2.16-1+ubuntu14.04.1+deb.sury.org+1MySQL Version: 5.5.5Caldera Forms Version: 1.8.2WP_DEBUG: 1

## Does Your Issue Persist When You Disable All Other Plugins and Switch To The Default Theme?
Yes

## What Is The Unexpected Behaviour?
When adding a Caldera Forms Gutenberg block and filling in a class name under the advanced block options ("Additional CSS Class"), the class name is not added to the containing div when rendered on the front-end.

## What PHP Errors Have You Logged While Reproducing This Bug?
None

## What JavaScript Errors Have You Seen While Reproducing This Bug?
None

tmdesigned:
PR #3092

PR #3092, opened 2019-03-14
#3091 - Value of additional CSS class name in Caldera Forms block added to rendering

tmdesigned:
The value of the "Additional CSS Class" field was already being passed in the $atts array to the rendering function. If a class name is present, I am utilizing the existing "caldera_forms_render_form_wrapper_classes" hook to add the class name. 

This shows what I think is the minimal solution. It may be better abstracted into 1-2 functions for cleanliness.

#3091


tmdesigned:
Unrelated, if they're bothersome, you can tell me how to avoid the extraneous commits that come from keeping my fork up to date.

Shelob9:
@tmdesigned Thanks for sending this.

Quick question: does this change which element gets that class added or duplicate the number of times it is added? If so, that's a backwards-compat breaking change in terms of how CSS is parsed and JavaScript bindings happen.

> If a class name is present, I am utilizing the existing "caldera_forms_render_form_wrapper_classes" hook to add the class name.

It is OK by WordPress core development standards to duplicate a hook via cut and paste. I disagree, I would prefer one method, wrapping the filter, used in both places.

>Unrelated, if they're bothersome, you can tell me how to avoid the extraneous commits that come from keeping my fork up to date.
This is because you are modifying your develop branch, instead of making a branch from develop for this change. You can read our complete workflow, and why we follow it here:

https://github.com/CalderaWP/Caldera-Forms/blob/master/contributing/pull-request-workflow.md

tmdesigned:
No problem.

Don't know if I understand your first question about changing which element gets the class. Right now _nothing_ gets the class. You can type a value in the block sidebar, and the existing render function receives it but ignores it. This adds it to the output of the topmost level of the block, which is where it goes for other Gutenberg blocks, and should only apply it once.

I don't think I explained clearly about the hook. I am not duplicating the filter, I am tying into it (i.e. add_filter) so as not to duplicate code. I just meant to avoid an anonymous function with the closure we could abstract the add_filter into another function.

Thanks for the github workflow tip.

Shelob9:
>Don't know if I understand your first question about changing which element gets the class. Right now nothing gets the class. 

I didn't understand you. This makes sense. I'm adding to 1.8.4 milestone, we can't scope creep 1.8.3.

> I just meant to avoid an anonymous function with the closure we could abstract the add_filter into another function.
I have no problem with an anonymous function on principle and I don't see why it needs to be a named function -- I would not want this to be removable and I don't see why using the function directly is helpful.

Issue #3065, opened 2019-03-08
Front end page load size increased


Shelob9:
Reported here https://github.com/CalderaWP/Caldera-Forms/pull/3052#issuecomment-470824631

vaka2vaka:
One page
Caldera 1.7.6 page size 19,14 KB
Caldera 1.8.2 Page Size 28.84 KB

Shelob9:
We need to audit if we're loading uneeded coded in new webpack bundle or enqueuing too many WordPress dependencies or both. Also, can we start removing cf1 files (ajax core, conditionals, etc.) in 1.8.3?

Shelob9:
Yah, we're bundling dependencies that are redundant to WordPress. This can be fixed in webpack config.

Shelob9:
Realted: https://stackoverflow.com/questions/33154285/exclude-react-from-webpack-bundle/41378514#41378514

As of ce58879 ~100k reduced

vaka2vaka:
This is a template problem.
Problem in GeneratePress
In OceanWP the problem disappears.
Thank

ictbeheer:
I don't think it is purely a theme problem. On a testsite with WP 5.1 and only Caldera Forms (and Classic Editor) activated the number of enqueued JS-files increases dramatically between CF 1.8.1 and 1.8.2:
1.8.1:  9 JS-files /  550 KB
1.8.2:  47 JS-files / 3.4 MB

vaka2vaka:
**ictbeheer**  Yes you are right.   CF1.8.2 creates problems.

salihkulangara:
This issue seems not related to any specific theme/template, I am using different themes for my clients, after the latest update of Caldera forms facing this issue, load a lot of js from wp-includes mostly related to react(seems related to editor). This heavily impact on the Performance, so for now I had to roll back to the version 1.7.7

Shelob9:
New PR #3113

Shelob9:
In #3126 we found that the blocks client was loading when it was not needed. That has a lot of extra dependencies, so that fix in 1.8.4 should help.

PR #3113, opened 2019-03-21
Reduce bundle size by using WordPress' react/ react DOM

Shelob9:
Excludes React from build - 1a6e70e - and uses WordPress' instead - 217f357

 #3065


New0:
@Shelob9 this seems to break CF2 fields on WP anterior of 5.0 ( worked on #3100 for 1.8.3 )

Issue #3157, opened 2019-04-10
Text between two % characters


teamtacho:
# PLEASE READ THIS LINE & Then Delete - We need this information to best understand your issue, please answer all questions!

## Do You Need Immediate Help?
yes

## Is This A Feature Request?
no

## What Version Of Caldera Forms, WordPress and PHP Are You Using?
WordPress Version: 4.9.8PHP Version: 7.2.15MySQL Version: 5.6.42Caldera Forms Version: 1.8.4

## What Is The Unexpected Behaviour?
If a text in a textarea field contains two % characters (e.g. ..the incline is between 20% and 30%..)
the % and the text between the % is not shown in the email. Obviously CF interprets this as a slug. The full text is shown in the CF entry viewer but not in the email.

## What PHP Errors Have You Logged While Reproducing This Bug?
-

## What JavaScript Errors Have You Seen While Reproducing This Bug?
-

teamtacho:
Hi
I have updated CF to version 1.8.5 but it isn't solved yet. Please can you give me guidance if I can do something or re-open that task fix the bug in the upcomming version.
Best Regards
Michael

New0:
Closed via - https://github.com/CalderaWP/Caldera-Forms/commit/a59c771c46e6973fabebeb958a315c17e75722eb -

PR #3168, opened 2019-04-16
Remove the magic tags from the options label

### LINKED ISSUE #3167: PHP Fatal error: Uncaught TypeError: Argument 2 passed to Caldera_Forms_Field_Util::has_field_type() must be of the type array, null given
On the default install of caldera if I try to view an entry it returns the following PHP Error
PHP Fatal error: Uncaught TypeError: Argument 2 passed to Caldera_Forms_Field_Util::has_field_type() must be of the type array, null given

It happens when there is something url encoded in the label of a an option (Eg: %C3%C3) (If you put a link inside the label for example)
That's because the option label tries to apply magic tags when it shouldn't

I tracked it down to the method format_select_options in core.php and the call to `self::do_magic_tags($option['label']);`

Why is the label magic tagged? There is no point in it is there? Anyway even if there was it would cause an error when viewing the entry because the form global is not defined at this point and would throw this error

Tofandel:
Fixes #3167


Tofandel:
> I can imagine scenarios where magic tags are used in options labels or values, like getting data from a previous page

The point is, it's not possible, if that was the case you'd get a fatal php error ;)

> Also, I'm not sure why you want to use an url in url encoded format in the option label but in this case I would first try using the 'caldera_forms_render_get_field' filter hook to format the option label as needed, you can read more about that hook at https://calderaforms.com/doc/caldera_forms_render_get_field/

It's quite a frequent use case, you can insert html in the label so if you have an <a> tag with a link in the label you will run into the error

Tofandel:
Related https://github.com/CalderaWP/Caldera-Forms/issues/3157

New0:
@Tofandel Thank you for the explanation, I missed the point that it was related to the entry viewer. I opened #3207 that address that issue. Magic tags in the option label working on frontend, I still believe we can't just remove them.

Also in the case you use a link in the option label, I'm not sure you need to url encode it.

Tofandel:
You don't always have control over the url, for me it's my client who is usually creating them, he his just copy pasting the url and putting it in the `<a href="url">` since it's in an html parameter and if there is special characters in it (like quotes or spaces) it has to be url encoded, also when you copy it from chrome you get it url encoded anyway

Shelob9:
>You don't always have control over the url, for me it's my client who is usually creating them, he his just copy pasting the url and putting it in the <a href="url"> since it's in an html parameter and if there is special characters in it (like quotes or spaces) it has to be url encoded, also when you copy it from chrome you get it url encoded anyway

In my opinion, this is not something we should support via magic tags. It's an edge case that's hard to control for securely. We'd have to detect if it is a link and if so what steps were need to escape it properly. I think using the `caldera_forms_render_get_field` to solve that specific issue makes more sense.

Something like:

```
add_filter( 'caldera_forms_render_get_field', function( $field )  {
    if( 'fld...' == $field[ 'ID' ] ){
        $url = function_that_determines_url_you_need();
        if( filter_var($url, FILTER_VALIDATE_URL ) ){
            $field[ 'config' ][ 'label' ] = '<a href="' . esc_url( urlencode($url ) ) . '">Text</a>';
            
        }
    }
    return $field;
});
```

https://calderaforms.com/doc/caldera_forms_render_get_field/

Shelob9:
@new0 Should we close this now that  #3207 is merged?

Issue #2335, opened 2018-03-01
create_function() is deprecated in php 7.2


Shelob9:
We use this for calculation fields because we can't use closures and support PHP 5.2.

I am not maintaining backwards compat with PHP 5.2 at the expense of users on PHP 7.2. You've had a decade to update, I'm sorry. Caldera Forms 1.7.0 -- IE the next major release after the one we're working on now will require PHP 5.3 or later. PHP7 for 2.0 is obviously going to be the case.

http://php.net/manual/migration72.deprecated.php

TomasVotruba:
If you need to upgrade more than one `create_function` to anonymous function ([official recommended upgrade](http://php.net/manual/en/migration72.deprecated.php#migration72.deprecated.create_function-function)), you can use a CLI tool for that.

It's tested on [30 various (and really weird :)) cases](https://github.com/rectorphp/rector/tree/c94f3208c0275c66385931e497afaf1001dfcc1e/packages/Php/tests/Rector/FuncCall/CreateFunctionToAnonymousFunctionRector/Fixture).

```diff
-$callback = create_function('$a', 'return "<cas:proxy>$a</cas:proxy>";');    
+$callback = function ($a) {
+    return "<cas:proxy>{$a}</cas:proxy>";
+};
```

Includes concat (.), string quotes and inclined function calls:

```diff
-$func = create_function('$atts, $content = null','return "<div class=\"' . $class_list . '\">" . do_shortcode($content) . "</div>";' );
+$func = function ($atts, $content = null) use ($class_list) {
+    return "<div class=\"{$class_list}\">" . do_shortcode($content) . "</div>";
+};
```

Do you want to automate the hard work?


### 1. Instal Rector

```bash
composer require rector/rector --dev
```

### 2. Create config

```yaml
# rector.yml
services:
    Rector\Php\Rector\FuncCall\CreateFunctionToAnonymousFunctionRector: ~
```

### 3. Upgrade your Code


```bash
vendor/bin/rector process src --config rector.yml --dry-run
vendor/bin/rector process src --config rector.yml
```

Enjoy!

Shelob9:
PR #2902 should resolve this.

Shelob9:
Steps:
[...]

PR #3182, opened 2019-04-29
Unit Tests For Caldera_Forms::run_calculation()

New0:
This will replace deprecated create_function() by a math parser.

This is work for #2335

### DIFF
--- a/classes/core.php
+++ b/classes/core.php
@@ -1450,17 +1450,8 @@ class Caldera_Forms
 				__('Calculation is invalid (division by zero)', 'caldera-forms'));
 		}
 
-		/**
-		 *  Set result out of formula Parser 
-		 *
-		 * @since 1.8.5
-		 * 
-		 * @param $formula 
-		 * @param $precision non required, Number of digits after the decimal point
-		 */
-		$parser = new FormulaParser($formula);
-		$result = $parser->getResult();
-		$total = $result[1];
+		$total_function = create_function(null, 'return ' . $formula . ';');
+		$total = $total_function();
 
 		if (is_infinite($total) || !is_numeric($total)) {
 			return new WP_Error($field['ID'] . '-calculation', __('Calculation is invalid', 'caldera-forms'));
--- a/classes/core.php
+++ b/classes/core.php
@@ -14,7 +14,12 @@
  * Caldera_Forms Plugin class.
  * @package Caldera_Forms
  * @author  David Cramer <david@digilab.co.za>
+ * 
+ * 
  */
+
+use FormulaParser\FormulaParser;
+
 class Caldera_Forms
 {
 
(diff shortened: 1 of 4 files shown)


tests/test-run-calculation.php
```diff
@@ -0,0 +1,273 @@
...
+            'fld_93143' => '1:1',
+            'fld_7465068' => '1:1',
+            'fld_8568604' => '1:1',
+            'fld_3665187' => '1:1',
+          ),
+          'structure' => '12',
+        ),
+        'fields' => 
+        array(
+          'fld_1892141' =>
```
Shelob9: @New0 You have some of the fields in the form config using fld_... for ID and some using a meaningful ID. Let's make them all meaningful and mayb use class properties. `$this->IdOdFieldThatHasValueOf1;  $this->manualCalcFieldId`

tests/test-run-calculation.php
```diff
@@ -0,0 +1,273 @@
...
+        $field = $form['fields']['fld_8568604'];
+
+        Caldera_Forms::set_field_data('fld_1892141', $fieldOneValue, $form );
+        Caldera_Forms::set_field_data('fld_93143', $fieldTwoValue, $form );
+        Caldera_Forms::set_field_data('fld_7465068', $fieldTwoValue, $form );
+        
+        $result = ($fieldOneValue + $fieldTwoValue) / $fieldThreeValue;
+
+        //Test Calculation by fields
+        $cf_calc = Caldera_Forms::run_calculation( null, $field, $form );
```
Shelob9: @New0 Small point, but changed `$cf_calc` to `$expectedResult` that would be more obvious when reviewing the test later.

New0:
Thanks for your comments @Shelob9 , it helped me realize a typo was making the test to fail

New0:
( @Shelob9 run composer install to test this )

classes/core.php
```diff
@@ -1446,8 +1450,17 @@ static public function run_calculation($value, $field, $form)
...
-		$total = $total_function();
+		/**
+		 *  Set result out of formula Parser 
+		 *
+		 * @since 1.8.5
+		 * 
+		 * @param $formula 
+		 * @param $precision non required, Number of digits after the decimal point
+		 */
+		$parser = new FormulaParser($formula);
```
Shelob9: @New0 Let's revert this until we have all of the tests in place. Once they pass as-is, then we swap out the parser.

Shelob9:
As we discussed yesterday, this PR will NOT fix #2335 But the tests that @New0 wrote are important  valuable, no matter how we solve #2335 and also useful to have. So this PR is now for #3203

Shelob9:
@New0 I think that this PR is blocked by:

- [ ] Need to reset to c837f5c
- [ ] Remove parser as dependency

Right? If so, please do and then request review again.

New0:
@Shelob9 correct for the Math parser, it was still listed in composer requirements. For the eval() function, it was already reverted and the core.php file is back to deprecated create_function() function

PR #3197, opened 2019-05-13
Removed git merge stuff in comment

ideadude:


Shelob9:
Thanks Jason! @New0 Editing this against dev gets messy. Probably going to need to grab this manually when we merge to master, but its worth fixing.

Issue #3141, opened 2019-04-03
1.8.4 - Reference Error - wp is not defined


colbyalbo:
after updating to 1.8.4, the console logs this Reference error.

Reference Error - wp is not defined, and points to file " Caldera-forms/clients/render/build/index.min.js

No visible effect on the front end that i can tell

screenshots attached
![caldera1](https://user-images.githubusercontent.com/30237541/55490057-b517d880-55f8-11e9-9edc-3706f77d65d5.jpg)
![caldera2](https://user-images.githubusercontent.com/30237541/55490079-c234c780-55f8-11e9-905a-83864b561a36.jpg)

colbyalbo:
upon further testing, it seems this is preventing the form from submitting.  I'll post the JS console errors later

salih-cairs:
While trying to fix this, I see when we load wp-embed scripts, the error goes and forms working back again. But it seems not ok, because most of the time to reduce the JS load size we are de-registering the wp-embed  scripts, right?

salihkulangara:
While trying to fix this, I see when we load wp-embed scripts, the error goes and forms working back again. But it seems not ok, because most of the time to reduce the JS load size we are de-registering the wp-embed scripts, right?

New0:
Hello @colbyalbo , could you share the information asked in the template ? 

- share the debug information from your WordPress admin at Caldera Forms -> Support under the debug information tab
- report if the issue persists with all other plugins disabled and theme twentynineteen activated.
- share PHP and JS logs

Another user reported this issue and had it fixed by deactivating and reactivating plugins, #3144

@salihkulangara thank you for investigating this, we fixed a similar issue in previous version, what happened was that wp was not loaded if WordPress version was older than 5.0.0, we are now loading to prevent this issue. Loading wp-embed script probably enqueued wp, which solved the issue on your installation.

Also be sure to have all caches fully flushed to test that issue.

Shelob9:
Related 
https://github.com/wp-media/wp-rocket/issues/1561
 #3144

Shelob9:
In this helpscout ticket - https://secure.helpscout.net/conversation/830660038/20516/ We see this error. An optimization plugin of some sort is being used here. I will ask which one. 97% chance this is an issue with an "optimization"

![Screen Shot 2019-04-19 at 9 04 07 AM](https://user-images.githubusercontent.com/1994311/56425403-1fc84580-6282-11e9-9094-46d68598793f.png)

[...]

Issue #3132, opened 2019-03-29
1.8.4 Uncaught TypeError


vaka2vaka:
Errors appear on two sites.
At first site -  when adding a file upload v2
When this file upload v2 appears in the designer, it is not reflected on the page.
Uncaught TypeError: Cannot read property 'fields' of undefined
    at index.min.js?ver=1.8.4:30
    at NodeList.forEach (<anonymous>)
    at HTMLDocument.<anonymous> (index.min.js?ver=1.8.4:30)
    at HTMLDocument.dispatch (jquery.js?ver=1.12.4:3)
    at HTMLDocument.r.handle (jquery.js?ver=1.12.4:3)
    at Object.trigger (jquery.js?ver=1.12.4:3)
    at Object.a.event.trigger (jquery-migrate.min.js?ver=1.4.1:2)
    at HTMLDocument.<anonymous> (jquery.js?ver=1.12.4:3)
    at Function.each (jquery.js?ver=1.12.4:2)
    at a.fn.init.each (jquery.js?ver=1.12.4:2)

In the second site
(When activated - Activate Ajax sending, the error disappears)

Uncaught TypeError: a(...).baldrick is not a function
    at resBaldrickTriggers (caldera-forms-front.min.js?ver=1.8.4:1)
    at HTMLDocument.<anonymous> (caldera-forms-front.min.js?ver=1.8.4:1)
    at i (jquery.js?ver=1.12.4:2)
    at Object.fireWith [as resolveWith] (jquery.js?ver=1.12.4:2)
    at Function.ready (jquery.js?ver=1.12.4:2)
    at HTMLDocument.K (jquery.js?ver=1.12.4:2)

Debug:
Uncaught TypeError: a(...).baldrick is not a function
http:/.................../wp-content/plugins/caldera-forms/assets/build/js/caldera-forms-front.min.js?ver=1.8.4 line 1

regiside:
I'm getting this error as well when I disable "Enable Ajax Submissions. (No page reloads)"
Uncaught TypeError: a(...).baldrick is not a function

Seems like forms are still able to be submitted though.

------Full Debug Information----
Wordpress Version 5.2
Full Debug Information
WordPress Version: 5.2

PHP Version: 7.2.17

MySQL Version: 5.5.5

Server Software: Apache/2.4.39 (Unix) OpenSSL/1.1.0g

[...]

PR #3202, opened 2019-05-16
Put a real big warning on forms that wp isn't defined and form will not work

### LINKED ISSUE #3199: wp is not defined should be caught and show a notice
Show a notice when WP is not defined that explains the known issues

Shelob9:
Fixes #3199

This shows a notice above the form if `window.wp` is not defined, which it may not be - #3141 
Can be cacused by https://fr.wordpress.org/plugins/disable-embeds/ or WP rocket before 3.3.3 https://github.com/wp-media/wp-rocket/issues/1720

I am considering removing the form if this fails, as its not going to work and if `wp` isn't defined, Caldera Forms can't be expected to function properly.

👀 Right now this is the alert ->
Screen Shot 2019-05-16 at 12 05 58 PM

That language is not great. I will think on it and ask WP Buffs folks what they think.


classes/field/localizer.php
```diff
@@ -104,14 +104,21 @@ public static function localize_cb(){
...
 	protected static function error_strings(){
-		$strings = array(
-			'mixed_protocol' => __( 'Submission URL and current URL protocols do not match. Form may not function properly.', 'caldera-forms' ),
-			'jquery_old'     => __( 'An out of date version of jQuery is loaded on the page. Form may not function properly.', 'caldera-forms' )
-		);
-
-		if( ! current_user_can( Caldera_Forms::get_manage_cap( 'admin' )) ){
-			unset( $strings[ 'jquery_old' ] );
-		}
+        if( current_user_can( Caldera_Forms::get_manage_cap('admin') ) ) {
```
Shelob9: @New0 This is a bit of a change - less info for people who can't fix the problem, more info for people who can.

classes/field/localizer.php
```diff
@@ -104,14 +104,21 @@ public static function localize_cb(){
...
+                'wp_not_defined' => __( 'The variable wp is not defined, your forms will not work.', 'caldera-forms' ),
+                'mixed_protocol' => __( 'Submission URL and current URL protocols do not match. Form may not function properly.', 'caldera-forms' ),
+                'jquery_old'     => __( 'An out of date version of jQuery is loaded on the page. Form may not function properly.', 'caldera-forms' )
+            );
+        }else{
+            $message =  __( 'Form may not function properly. Contact site administrator.', 'caldera-forms' );
+            $strings = array(
+                'wp_not_defined' => $message,
+                'mixed_protocol' => $message,
+                //jquery_old is intentionally omitted. See assets/js/front-end-script-init.js ~L378.
```
Shelob9: LOL...

Tabrisrp:
@Shelob9 do you understand why removing the inline emoji script would cause this error in Caldera?

Shelob9:
It causes the `wp` global to be undefined. That is bad as we rely on `wp.element`, which we enqueue using the standard APIs. I don't understand why that happens. I presume that core is pressuming that `wp` is going to be defined when it starts attaching stuff to it in `wp-includes/script-loader.php` Look in `wp_default_scripts()`. How are you removing emojis?

Tabrisrp:
Everything is in there: https://github.com/wp-media/wp-rocket/blob/master/inc/common/emoji.php

When @webtraingwheels tested, she found that it was specifically this line that was causing the issue: https://github.com/wp-media/wp-rocket/blob/master/inc/common/emoji.php#L11

I really fail to see which of these is related to the `wp` global unfortunately.

Shelob9:
@Tabrisrp OK, this makes a lot more sense after seeing what WP Rocket is doing. Also WordPress and/ or Josh are silly.

So, one of the things that disabling emojis is doing is preventing this line of code from running 
* https://github.com/WordPress/WordPress/blob/master/wp-includes/js/wp-emoji.js#L290-L295

That sets up the `wp` variable. We are enqueuing wp-element using a lot of logic on top of normal wp_enqueue/register_script ->
* https://github.com/CalderaWP/Caldera-Forms/blob/master/classes/render/assets.php#L647
* https://github.com/CalderaWP/Caldera-Forms/blob/master/classes/render/assets.php#L665

Are webpack config treats WordPress dependencies as a externals and doesn't bundle them. It assumes that `wp.element` is set. That apparently is not be possible, if core didn't do `window.wp = window.wp || {};`

Tabrisrp:
Ok so what we can do on our side is, if we detect that Caldera forms is active, we prevent the use of the disable emoji option, adding an explanation in our settings page. Looks good for you?

Shelob9:
That would work, I wonder if there is a simpler solution. I also worry about how this effects WP Rocket + other plugins that use WordPress core scripts in front-end. Maybe printing `window.wp = window.wp || {};` in the header where emoji loader would go?

Shelob9:
Also fixes #3132

Shelob9:
Also, fixes #3141

PR #3207, opened 2019-05-20
If $form is null in Caldera_Forms_Magic::field_magic set it to be an …

### LINKED ISSUE #3167: PHP Fatal error: Uncaught TypeError: Argument 2 passed to Caldera_Forms_Field_Util::has_field_type() must be of the type array, null given

New0:
…empty array.

That prevents option labels that use magic tags to break entry viewer, when $form global is not defined on the admin side

Fixes #3167


Shelob9:
@New0 Quick reminder -> PRs to fix bugs should have a form export attached.

PR #3208, opened 2019-05-20
Work on calculation field for PHP version >= 7.2.0

New0:
- use parser if php version is >= 7.2.0 ( => parser is limited to sqrt, abs, sin, cos, tan, log, exp functions and don't support , and ** )
- If using parser, show errors for deprecated symbols and function

This is work for #2335


Shelob9:
Testing and works as expected (I see same total in front-end preview and entry viewer) with test forms:

-  1979 Calculations Created@1.5.5 dev@1.5.6.2
- 1962 Calculations created@1.5.5 dev@1.5.6.2
- 1993 Calculations created@1.5.5 dev@1.5.6.2
- X1 Calculations create@1.5.5 dev@1.5.6.2
- X2 Calculations create@1.5.5 dev@1.5.6.2
- X3 Calculations create@1.5.5 dev@1.5.6.2  - manual formula `round( tan( %number% ) )`
- 2063 Calculations create@1.6.2.2 dev@1.5.7.1
- 2063.2 Calculations create@1.6.2.2 dev@1.5.7.1

Shelob9:
Also tested with test forms:

- Calculation Money

Shelob9:
Ok, so I think this works, but we need to work on the notices.

classes/core.php
```diff
@@ -1447,9 +1470,40 @@ static public function run_calculation($value, $field, $form)
...
 				__('Calculation is invalid (division by zero)', 'caldera-forms'));
 		}
+	
+		//If PHP version is less than 7.2.0, continue using old function
+		if(phpversion() < '7.2.0'){
+
+			$total_function = create_function(null, 'return ' . $formula . ';');
+			$total = $total_function();
+
+		} else { //avoid using create_function() when using PHP version >= 7.2.0
```
Shelob9: comments should be on a separate line.

classes/core.php
```diff
@@ -1447,9 +1470,40 @@ static public function run_calculation($value, $field, $form)
...
+
+			//Check if the formula use a symbol or function not supported by the parser but declared as supported at https://calderaforms.com/doc/calculation-fields/
```
Shelob9: @New0 All of this inline code explaining WHY you did the logic in this order is great. Helps with code review and I'm sure it will help us later.

classes/core.php
```diff
@@ -1447,9 +1470,40 @@ static public function run_calculation($value, $field, $form)
...
+				$parser = new FormulaParser($formula);
+				$result = $parser->getResult();
+				//Get result if parser returns a correct status
+				if($result[0] === "done"){ 
+					$total = $result['1'];
+				} else if($result[0] === "error") {  //Return an error if parser returned an error status
+					return new WP_Error($field['ID'] . '-calculation', __('Deprecated operator or symbol in calculation for PHP version >= 7.2 requirements', 'caldera-forms'));
+				}
+				
+			} else { //Return an error with the deprecated symbol or function caught
```
Shelob9: @New0 I don't agree with this decision. I think we should still run the submission. If they are an admin, we should attach this warning to the form notices. If not, report nothing.

Check out here where I did the notices differnetly based on if the current user can fix the issue or not:

https://github.com/CalderaWP/Caldera-Forms/pull/3202/files#diff-cda146b98bb4831185232d8634a1f85eR109

See here for example of changing admin notices. I suspect this will require its own class
https://github.com/CalderaWP/Caldera-Forms/blob/master/includes/cf-pro-client/classes/hooks.php#L37-L38

classes/core.php
```diff
@@ -1447,9 +1470,40 @@ static public function run_calculation($value, $field, $form)
...
+				$message = sprintf(
+					'"%s" %s',
+					$deprecated,
+					__('has been deprecated in calculations for PHP version >= 7.2', 'caldera-forms')
```
Shelob9: Please make the message: `%s has been deprecated in calculation fields for compatibility with PHP  7.2 or later`

classes/core.php
```diff
@@ -1390,6 +1392,85 @@ static public function akismet_scanner($config, $form)
...
+					'"%s" %s',
+					$deprecated,
+					__('has been deprecated in calculation fields for compatibility with PHP 7.2 or later', 'caldera-forms')
+				);
+
+				$out['error']['note'] = $message;
+
+				return $out;
+
+			}, 10);
```
Shelob9: Arguably, this should have a higher priority. We use 10 in a few places https://github.com/CalderaWP/Caldera-Forms/search?q=caldera_forms_ajax_return&unscoped_q=caldera_forms_ajax_return

Moving it to 11 would make it over ride CF Pro notices, which is good. So maybe set to 25, but not 99.

classes/core.php
```diff
@@ -1390,6 +1392,85 @@ static public function akismet_scanner($config, $form)
...
+	 * @param string $formula to be sent to original calculation job
+	 * @param string $deprecated function or symbol caught
+	 * 
+	 * @return original process and notice for admin users
+	 */
+	static public function calculation_deprecated_caught_process( $formula, $deprecated )
+	{
+		if( current_user_can( Caldera_Forms::get_manage_cap('admin') ) ) {
+			
+			add_filter('caldera_forms_render_notices', function( $out ) use ($deprecated){
```
Shelob9: Does this work with AJAX submissions on? I __think__ you have to use caldera_forms_ajax_return for that, I could be wrong.

classes/core.php
```diff
@@ -1447,9 +1528,37 @@ static public function run_calculation($value, $field, $form)
 			return new WP_Error($field['ID'] . '-calculation',
 				__('Calculation is invalid (division by zero)', 'caldera-forms'));
 		}
+	
+		//If PHP version is less than 7.2.0, continue using old function
+		if(phpversion() < '7.2.0'){
```
Shelob9: @New0 Please use version compare here. This code probably works as expected, but it's unreliable,  `version_compare` solves that.

https://www.php.net/manual/en/function.version-compare.php

classes/core.php
New0: @Shelob9 it works with ajax submissions but not without... And I noticed (requires further tests) the PDF link only appears with non-ajax submissions

PR #3219, opened 2019-05-23
Put a real big warning on forms that wp isn't defined and form will not work (2)

New0:
Rename wpDefined to wpUndefined (as true means wp is not defined) and change the condition that triggers the warning.

This replaces #3202 that has been merged too fast...

### DIFF
--- a/assets/js/frontend-script-init.js
+++ b/assets/js/frontend-script-init.js
@@ -345,8 +345,7 @@ var cf_jsfields_init, cf_presubmit;
 window.addEventListener("load", function(){
 	(function( $ ) {
 		'use strict';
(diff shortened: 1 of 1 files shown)


assets/js/frontend-script-init.js
```diff
@@ -345,6 +345,8 @@ var cf_jsfields_init, cf_presubmit;
 window.addEventListener("load", function(){
 	(function( $ ) {
 		'use strict';
+    //Catch if window.wp is undefined
+		var wpUndefined = undefined === typeof window.wp;
```
Shelob9: Changing the name to match conditional's purpose makes a lot of sense here.

Issue #3228, opened 2019-05-24
Add round support for calculation field when PHP >= 7.2


New0:
The formula parser used don't support round, switching to math parser will cover this and ceil and floor.

Shelob9:
Close via #3229

PR #3229, opened 2019-05-24
Add support for round, ceil and floor on calculation field when using…

New0:
… PHP 7.2

Switched from formula parser to math parser ( https://github.com/mossadal/math-parser )

For #3228

### DIFF
--- a/classes/core.php
+++ b/classes/core.php
@@ -18,7 +18,8 @@
  * 
  */
 
-use FormulaParser\FormulaParser;
+use MathParser\StdMathParser;
+use MathParser\Interpreting\Evaluator;
 
 class Caldera_Forms
 {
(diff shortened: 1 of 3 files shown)


classes/core.php
```diff
@@ -1404,7 +1405,7 @@ static public function akismet_scanner($config, $form)
 	static public function check_deprecated_math_functions($formula)
 	{	
 		//List known depracated functions
-		$deprecated = array( ',', '**', 'pow', 'acos', 'asin', 'atan', 'atan2', 'ceil', 'floor', 'max', 'min', 'random', 'round');
+		$deprecated = array( ',', '**', 'pow', 'acos', 'asin', 'atan', 'atan2', 'max', 'min', 'random');
```
Shelob9: @New0 arrays should use square brackets and have one entry per line.

Issue #3232, opened 2019-05-28
Run tests in WordPress 4.9, 5.0, 5.1 and latest


Shelob9:
Currently we test latest and real old versions we should test 4.9 and later.

Related  #3205

New0:
Closed by https://github.com/CalderaWP/Caldera-Forms/commit/c3c020ab2bece40572fa084c3e3ee35d086222a0

PR #3233, opened 2019-05-28
Run tests in WordPress 4.9, 5.0, 5.1 and latest

Shelob9:
#3232

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,8 @@ sudo: false
 language: php
 
 php:
+  - '7.3'
+  - '7.2'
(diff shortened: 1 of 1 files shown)


Shelob9:
Before: 
Screen Shot 2019-05-28 at 9 45 56 AM

After:
Screen Shot 2019-05-28 at 9 46 38 AM

Before, we had 9 runs, now 10. Now testing Php 5.6,7.0-3 with WP latest, then test JS and multisite with PHP 7.2 and test WP 4.9,5.0 and 5.1 with PHP 7.2.

PR #3264, opened 2019-06-14
First step of preparations + fixes

szepeviktor:
Hello @Shelob9 !

This is the first stage of discovering bugs by @phpstan

It works like this: use strict types, define properties before usage, fix invalid PHPDoc blocks, run PHPStan -> Benefit!

Some basic errors are fixed in this PR.

What do you think?


szepeviktor:
1 more thing, before running PHPStan you have to remove is_countable() in vendor/giacocorsiglia/wordpress-stubs/wordpress-stubs.php:73532 - it is a pending bug

szepeviktor:
Oh no! PHPStan should be installed manually: it needs modern PHP https://travis-ci.org/CalderaWP/Caldera-Forms/jobs/545600975

```
composer require --dev phpstan/phpstan-shim szepeviktor/phpstan-wordpress
```

szepeviktor:
Do you know cep_get_easy_pod() and cep_get_registry() ?

szepeviktor:
@New0 What do you think?

Issue #3262, opened 2019-06-13
window.cf2 is undefined in ajax-core.min.js


PASAf:
Maybe related to #2821, #2970, #3098 
WordPress Version: 5.2.1 Caldera Forms Version: 1.8.5

Error occurs in ajax-core.js when using Caldera Easy Pods forms with ajax

![image](https://user-images.githubusercontent.com/1716683/59414584-7063a700-8dca-11e9-9158-5f58ac9d7d53.png)
![image](https://user-images.githubusercontent.com/1716683/59414429-2b3f7500-8dca-11e9-8d1f-332edf13a217.png)

PASAf:
Please review my fix: https://github.com/CalderaWP/Caldera-Forms/pull/3265/commits/5aa1886cb767316c870771e02b9b20d0689c178c

Shelob9:
@PASAf Thanks for opening this issue. We would be happy to review a pull request. Your code looks good, but its hard to test with out a PR.

Here is the documentation on our PR workflow: https://github.com/CalderaWP/Caldera-Forms/blob/master/contributing/pull-request-workflow.md

Shelob9:
Related #3141

PASAf:
@Shelob9 I'm sorry, but I'm not able to do all those manipulations with git described in workflow now (create separate branch etc.)

Could someone reimplement this with following all rules?

PASAf:
@Shelob9 Question also is how to correctly update ajax-core.min.js?

New0:
Thank you for the PR @PASAf ,The minified file can be generated running grunt but it is not needed to test or push in the PR. It will be automatically generated during build process.

New0:
Closed via 373b3f3

PR #3265, opened 2019-06-14
Fix window.cf2 is undefinded for issue #3262

PASAf:
Make sure window.cf2 is set before declaring cf2


New0:
Thank you for opening the PR @PASAf for #3262. I could test that easy pods ajax query worked as expected and that check hid the error from the console.

New0:
@PASAf I'm closing this PR as I realize it's opened against master, I reproduced it against develop at https://github.com/CalderaWP/Caldera-Forms/pull/3289

PR #3281, opened 2019-07-09
fix: allow uploaded files not in media library to be filtered to new directory

DevinWalker:
## PR Description

The following PR adds the `caldera_forms_upload_directory` to the first conditional in the `add_upload_filter()` method so that I can have files upload to a specific directory and _not_ be shown in the WP Media Library.


Shelob9:
@New0 Devin and I discussed this as solution to his problem - it looks like that filter changes the upload directory for all upload types, but it's only for media library, which is silly.

This fix is tested on their site by Devin.

New0:
Thank you for this Pull Request @DevinWalker 

@Shelob9 I would rename the filter to caldera_forms_private_upload_directory to differentiate it from the public filter used line 131. And I would add the $transient_id as parameter for caldera_forms_private_upload_directory filter as it can be used in the returned function. What to you think ?

Shelob9:
Changing the name of the filter eliminates a lot of potential problems.

New0:
Travis build passed ( https://travis-ci.org/CalderaWP/Caldera-Forms/builds/556763747 )

New0:
Thanks again @DevinWalker I'm merging this to be released in 1.8.6 update.

Shelob9:
Documentation: https://calderaforms.com/doc/caldera_forms_private_upload_directory/

regiside:
Does this private upload functionality have a default directory or will it not work at all without implementing this filter? It would be nice if you could set files to upload but not be added to the media library without needing to edit any files for people less competent with PHP.

Issue #3270, opened 2019-06-21
If there is a CR (LF) after a manual mode formula in a field, the PHP parser breaks.


sburlot:
### BUG
If there is a CR (LF) after a manual mode formula in a field, the PHP parser breaks:
ie  "`%total% + %total2%\n`" instead of "`%total% + %total2%`"
The `MathParser\StdMathParser->parse('(90) + (0)\n')` fails (see PHP error log)

**The Javascript calculation works fine: no error.**

trimming the formula in PHP should prevent this error

## What Version Of Caldera Forms, WordPress and PHP Are You Using?
WordPress Version: 5.2.2, PHP Version: 7.2.19-0ubuntu0.18.04.1, MySQL Version: 5.7.26, Caldera Forms Version: 1.8.5, WP_DEBUG:

## Does Your Issue Persist When You Disable All Other Plugins and Switch To The Default Theme?
Yes

## What Is The Unexpected Behaviour?
Submitting the form does nothing (spinner in the middle of the screen)

## What PHP Errors Have You Logged While Reproducing This Bug?
```
[21-Jun-2019 21:50:58 UTC] PHP Fatal error:  Uncaught Error: Call to a member function isTerminal() on null in [...]/wp-content/plugins/caldera-forms/vendor/mossadal/math-parser/src/MathParser/Parsing/Parser.php:166
Stack trace:
#0 [...]/wp-content/plugins/caldera-forms/vendor/mossadal/math-parser/src/MathParser/Parsing/Parser.php(124): MathParser\Parsing\Parser->shuntingYard(Array)
#1 [...]/wp-content/plugins/caldera-forms/vendor/mossadal/math-parser/src/MathParser/StdMathParser.php(55): MathParser\Parsing\Parser->parse(Array)
#2 [...]/wp-content/plugins/caldera-forms/classes/core.php(1560): MathParser\StdMathParser->parse('(90) + (0)\n')
#3 [...]/wp-includes/class-wp-hook.php(286): Caldera_Forms->run_calculation('90', Array, Array)
#4 [...]/wp-includes/plugin.php(208): WP_Hook->apply_filters('90', Array)
#5 [...]/wp-content/plugins/caldera-forms/classes/core.php(249 in [...]/wp-content/plugins/caldera-forms/vendor/mossadal/math-parser/src/MathParser/Parsing/Parser.php on line 166
```

## What JavaScript Errors Have You Seen While Reproducing This Bug?
```
[...]

PR #3283, opened 2019-07-11
Trim Manual calculations

New0:
Trim manual calculations before saving form and before sending formula to new parser.

For #3270

### DIFF
--- a/classes/forms.php
+++ b/classes/forms.php
@@ -481,7 +481,7 @@ class Caldera_Forms_Forms {
 	 * @return string|bool Form ID if updated, false if not
 	 */
 	public static function save_form( $data, $type = 'primary' ){
-
+		
 		
 		if(!empty($data['fields'])){
 			foreach($data['fields'] as &$field){
@@ -494,7 +494,7 @@ class Caldera_Forms_Forms {
 					}
 				// trim manual calculations
 				} else if(!empty($field["config"]["manual_formula"]) && is_string($field["config"]["manual_formula"])){
-					$field["config"]["manual_formula"] = str_replace( "\r", '', str_replace( "\n", '', str_replace( ' ', '', trim( $field["config"]["manual_formula"] ) ) ) );
+					$field["config"]["manual_formula"] = Caldera_Forms_Sanitize::finish_trim( Caldera_Forms_Sanitize::sanitize_header( $field["config"]["manual_formula"] ) );
 				}
 			}
 		}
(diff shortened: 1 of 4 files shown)


classes/forms.php
```diff
@@ -482,15 +482,19 @@ protected static function add_details( $forms ){
...
+				// option value labels
 				if(!empty($field['config']['option']) && is_array($field['config']['option'])){
 					foreach($field['config']['option'] as &$option){
 						if(!isset($option['value'])){
 							$option['value'] = $option['label'];
 						}
 					}
+				// trim manual calculations
+				} else if(!empty($field["config"]["manual_formula"]) && is_string($field["config"]["manual_formula"])){
+					$field["config"]["manual_formula"] = str_replace( "\r", '', str_replace( "\n", '', str_replace( ' ', '', trim( $field["config"]["manual_formula"] ) ) ) );
```
Shelob9: @New0 This is a lot, without a test.

Can you please add a method to `Caldera_Forms_Santize` class to do this. And test against possible inputs, so we know it is safe?

Also what about `$str = trim( str_replace( '<br>', '', nl2br($str)  ));` nl2br should convert all line breaks to html br tags, which we can then lift out.

classes/forms.php
```diff
@@ -481,16 +481,20 @@ protected static function add_details( $forms ){
...
+				// option value labels
 				if(!empty($field['config']['option']) && is_array($field['config']['option'])){
 					foreach($field['config']['option'] as &$option){
 						if(!isset($option['value'])){
 							$option['value'] = $option['label'];
 						}
 					}
+				// trim manual calculations
+				} else if(!empty($field["config"]["manual_formula"]) && is_string($field["config"]["manual_formula"])){
+					$field["config"]["manual_formula"] = Caldera_Forms_Sanitize::finish_trim( Caldera_Forms_Sanitize::sanitize_header( $field["config"]["manual_formula"] ) );
```
Shelob9: This is better 👍

tests/test-forms-api.php
```diff
@@ -205,5 +205,35 @@ public function testUpdateAndGet()
...
+      		'required' 		=>	1,
+			'config'		=>	[
+				'manual_formula'	=> ' 2 + 2\n'
+			]
+		];
+		Caldera_Forms_Forms::save_form($form);
+
+		$form = Caldera_Forms_Forms::get_form($form_id);
+
+		$this->assertSame( '2 + 2', $form['fields']['fld_60011111']['config']['manual_formula'] );
```
Shelob9: 🌮

New0:
Travis passed at https://travis-ci.org/CalderaWP/Caldera-Forms/builds/560540426 but doesn't seem to be  returning the status

New0:
Passing build after update https://travis-ci.org/CalderaWP/Caldera-Forms/builds/562101324

PR #3289, opened 2019-07-22
Fix window.cf2 is undefinded caused by ajax requests

New0:
Copy of PR https://github.com/CalderaWP/Caldera-Forms/pull/3265 bt @PASAf

For #3262 

Thank you again @PASAf, I reproduced the PR against develop branch, yours was made against master.

This prevent errors when ajax request are made, precise case of easy pods ajax request that logged the error.

### DIFF
--- a/assets/js/ajax-core.js
+++ b/assets/js/ajax-core.js
@@ -174,7 +174,9 @@ jQuery(function($){
                 });
 
 				//Check if any cf2 fields are blocking submit
-				var cf2 = window.cf2[ $form.attr( 'id' ) ];
+                if( 'object' === typeof  window.cf2 ){
+                    var cf2 = window.cf2[ $form.attr( 'id' ) ];
+                }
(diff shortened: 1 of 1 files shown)


New0:
Passing Tests at https://travis-ci.org/CalderaWP/Caldera-Forms/builds/562109368

assets/js/ajax-core.js
```diff
@@ -174,7 +174,9 @@ jQuery(function($){
                 });
 
 				//Check if any cf2 fields are blocking submit
-				var cf2 = window.cf2[ $form.attr( 'id' ) ];
+                if( 'object' === typeof  window.cf2 ){
```
Shelob9: @New0 Nitpick, I am about to approve this, but keep in mind arrays, sets and maps are all objects. So `let cf = []; typeof cf` would return "object". In this case, if its an array, then that's intentional misuse, we can't worry about.

Issue #3288, opened 2019-07-22
Email tab options when pro enhanced delivery is enabled


New0:
This is possibly related to issues like #3284.

The options don't look as they should when pro enhanced delivery is activated, the "From Email" field should not appear when enhanced delivery enabled and should look like ->

![Screenshot_2019-07-19 Caldera Forms Admin ‹ VS Lando Dev — WordPress](https://user-images.githubusercontent.com/16348693/61523690-272ff400-aa15-11e9-944c-a86bab5678fb.png)

New0:
Closed via 3e18974

PR #3291, opened 2019-07-23
 Set correct state for IN-FORM enhanced delivery option

New0:
For #3288 

In the form settings under the pro tab, the checkbox for enhanced delivery was set backwards.


Shelob9:
@new0 Looks good.

New0:
Passing test at https://travis-ci.org/CalderaWP/Caldera-Forms/builds/562514562

Issue #3295, opened 2019-07-24
Uncaught Error: vi is not available in the catalog


tuanbbhero:
Failed to load resource: the server responded with a status of 404 (Not Found)
mysite.com/wp-content/plugins/caldera-forms/assets/js/i18n/vi.js?ver=1.8.5

Uncaught Error: vi is not available in the catalog
    at m.setLocale (parsley.min.js?ver=1.8.5:2)
    at i.window.Parsley.<computed> [as setLocale] (parsley.min.js?ver=1.8.5:3)
    at (index):1325

I encountered this problem from previous versions, to the current version (when I use Vietnamese at Settings > General).

New0:
Hello @tuanbbhero , thank you for reporting this issue. This comes from the ParsleyJS library that is used for messages in the validation process. Unfortunately it has not been translated to Vietnamese.

In the case you wish to contribute you can on the original repo at https://github.com/guillaumepotier/Parsley.js/blob/master/CONTRIBUTING.md or here sharing a vi.js file formatted as the other files you will find in this repo at  [assets/js/i18n/](https://github.com/CalderaWP/Caldera-Forms/tree/master/assets/js/i18n).

In the case you wish to contribute for general Vietnamese translation of the plugin, you can use https://translate.wordpress.org/projects/wp-plugins/caldera-forms/

tuanbbhero:
@New0 Thanks. I'm not developer. 
So, to solve the problem, I have 2 ways
+ Translating the vi.js file then upload to [assets/js/i18n](https://github.com/CalderaWP/Caldera-Forms/tree/master/assets/js/i18n)
+ Or translate the plugin at Translate.WordPress.org (I'm translating from a few months ago, but it has a lot, so it will probably take some months to complete).

Is that right? Thank you very much <3

New0:
Hey @tuanbbhero , I realize my answer was misleading, I'm sorry.

To fix this precise error message shown in the console logs ( Uncaught Error: vi is not available in the catalog ) you need to add a ( valid ) vi.js file within the assets/js/i18n directory, that is correct.

The files within the assets/js/i18n directory hold the translations for forms validation messages, it is not so long and can be useful to your users if the form catches a validation issue before submission.

The link to Translate.WordPress.org was just to show the way to translate the plugin and is not related to the precise issue you shared here.

In the case you translate the vi.js, please share it to be in a future release for other Vietnamese users.

Shelob9:
[...]

PR #3297, opened 2019-07-25
Fallback to "en" locale if no file exists in Parsley i18n folder

New0:
For #3295 

Fallback to "en" if no file in assets/js/i18n/ folder matches the locale ( we need to enqueue a file to be sure setLocale() is used for Parsley and it doesn't eturn an error.

@Shelob9 do you have an idea how we could test this ? Consequences are a logged error if no locale was set for Parsley or a 404 response for a missing enqueued asset.


Shelob9:
Isn't there a filter in WordPress for default language?

Also, we really should have a French test site and a test site in a language we don't have translation files running the acceptance tests.

classes/render/assets.php
```diff
@@ -1081,6 +1080,10 @@ protected static function set_locale_code($locale)
 			$locale = strtolower(substr($locale, 0, 2));
 		} elseif (file_exists(CFCORE_PATH . 'assets/js/i18n/' . strtolower(substr($locale, 3)) . '.js')) {
 			$locale = strtolower(substr($locale, 3));
+		} else {
+			//No file is matching the locale in validation folder, 
+			//Fallback to english instead of not enqueueing for Parsley to be set a locale and not return an error
```
Shelob9: Me: It smells bad that there is nothing in the else, but nothing is wrong.

Also me: It smelled bad for a reason.

tests/test-locale-assets-enqueued.php
```diff
@@ -0,0 +1,50 @@
...
+     */
+    public function testKnownLocale()
+    {
+        
+        $locale = "fr";
+        $code =   Caldera_Forms_Render_Assets::set_locale_code($locale);
+        $validator_url = Caldera_Forms_Render_Assets::get_validator_locale_url($locale);
+
+        $this->assertSame( $code, "fr" );
+        $this->assertSame( $validator_url, "http://example.org/wp-content/plugins/app/assets/js/i18n/fr.js" );
```
Shelob9: @New0 I would be careful about assuming site_url. Maybe better to check if the string ends with `i18n/fr.js`.

tests/test-locale-assets-enqueued.php
```diff
@@ -0,0 +1,50 @@
...
+
+     /**
+      * Test a locale which corresponding file doesn't exist within assets/js/i18n/
+      * 
+      * @since 1.8.7
+      *
+      * @covers  \Caldera_Forms_Render_Assets::set_locale_code($locale)
+      * @covers  \Caldera_Forms_Render_Assets::get_validator_locale_url($locale)
+     */
+    public function testUnkKnownLocale()
```
Shelob9: @New0 Type in method name

tests/test-locale-assets-enqueued.php
```diff
@@ -0,0 +1,50 @@
...
+    {
+        
+        $locale = "bo";
+        $code =   Caldera_Forms_Render_Assets::set_locale_code($locale);
+        $validator_url = Caldera_Forms_Render_Assets::get_validator_locale_url($locale);
+
+        $this->assertSame( $code, "en" );
+        $this->assertSame( $validator_url, "http://example.org/wp-content/plugins/app/assets/js/i18n/en.js" );
```
Shelob9: @New0 Same point as in other test about only testing the end.

classes/render/assets.php
```diff
@@ -1049,14 +1049,14 @@ protected static function field_script_dependencies()
...
 	 * @return string
 	 */
-	protected static function get_validator_locale_url($locale)
+	public static function get_validator_locale_url($locale)
 	{
-		if(!empty($locale)){
-			$locale = self::set_locale_code($locale);
-		}
+		//Use correct formatted locale for file to be matched
+		$locale = self::set_locale_code($locale);
```
mrpritchett: Why are we removing a check here?

classes/render/assets.php
mrpritchett: Agree

classes/render/assets.php
New0: @mrpritchett The check was removed to always run set_locale_url() which sets a default $locale to "en" if no locale was set or no file matching local found

tests/test-locale-assets-enqueued.php
```diff
@@ -0,0 +1,50 @@
...
+        
+        $locale = "bo";
+        $code =   Caldera_Forms_Render_Assets::set_locale_code($locale);
+        $validator_url = Caldera_Forms_Render_Assets::get_validator_locale_url($locale);
+
+        $this->assertSame( $code, "en" );
+        $this->assertSame( substr($validator_url, -10), "i18n/en.js" );
+    }
+
+}
```
mrpritchett: New line.

mrpritchett:
Approved pending new line.

Issue #3299, opened 2019-07-29
Uncaught TypeError: When adding two forms to one page with the Advanced File Uploader field (2.0)


vaka2vaka:
## What Version Of Caldera Forms, WordPress and PHP Are You Using?
5.2.2	WP 	 
Memory limit:	512M	

Caldera 1 8 6 
PHP 7 2
## Does Your Issue Persist When You Disable All Other Plugins and Switch To The Default Theme?
Yes
## What Is The Unexpected Behaviour?
When adding two forms to one page with one of them containing the Advanced File Uploader (2.0) field, an error occurs
________________________
Uncaught TypeError: Cannot read property 'fields' of undefined
     at index.min.js? h = 1053749907 & ver = 1.8.6: 30
     at NodeList.forEach (<anonymous>)
     at HTMLDocument. <anonymous> (index.min.js? h = 1053749907 & ver = 1.8.6: 30)
     at HTMLDocument.dispatch (jquery.js? ver = 1.12.4-wp: 3)
     at HTMLDocument.r.handle (jquery.js? ver = 1.12.4-wp: 3)
     at Object.trigger (jquery.js? ver = 1.12.4-wp: 3)
     at Object.a.event.trigger (jquery-migrate.min.js? ver = 1.4.1: 2)
     at HTMLDocument. <anonymous> (jquery.js? ver = 1.12.4-wp: 3)
     at Function.each (jquery.js? ver = 1.12.4-wp: 2)
     at a.fn.init.each (jquery.js? ver = 1.12.4-wp: 2)
________________________
 
In this case, the change in the second form ceases to be preserved.
This is very problematic if you need to add an order form and a callback form on a single page.
Thanks for attention

![423](https://user-images.githubusercontent.com/48130019/62036137-e2b50d00-b1f9-11e9-9cb4-404ce239da0a.jpg)

New0:
Hello @vaka2vaka , thank you for reporting and detailing the issue, we need to add a variable verification at that stage.

New0:
Cloased via https://github.com/CalderaWP/Caldera-Forms/commit/b9723874099e053e87fbfbcd03f71c18f9388301

PR #3302, opened 2019-07-29
Check if formConfig exists before trying to retrieve a fieldConfig

New0:
** yarn run dev to test this **

For #3299 

formConfig may not exist if there are multiple forms on same page, we are checking if it exists before trying to get fieldConfig


clients/render/index.js
```diff
@@ -38,23 +38,27 @@ domReady(function () {
 		//Build configurations
 		document.querySelectorAll('.cf2-field-wrapper').forEach(function (fieldElement) {
 			const fieldIdAttr = fieldElement.getAttribute('data-field-id');
+			if( 'undefined' !== typeof idAttr ){
```
mrpritchett: Inconsistent spacing between `if` and `(` and the next char. It's different here and everywhere else in the file.

New0: Ok, I removed all spaces in condition statements

mrpritchett:
Approved.

PR #3306, opened 2019-07-30
Migrate To GH actions and Gitlab CI

### LINKED ISSUE #3152: Replace Travis With Github actions

Shelob9:
Fixes #3152

Replaces #3153


Shelob9:
@New0 Reporting from the bottom of the rabbit hole:

* Setting up PHP unit tests (no WordPress) is done in first two steps, works.
* Setting up and running JavaScript tests is done in second two steps, works.
* Running the WordPress test suite, I can't figure out. I thaught that I could just install docker-compose and run it. See my last 27 commits.

I give up for the day/ ever. Why can't we just do this: https://git.saturdaydrive.io/_/ninja-forms/ninja-forms/blob/develop/.gitlab-ci.yml#L25-34 ?

Shelob9:
Seriously, we don't even have to migrate to Gitlab. You can run Gitlab CI for a Github repo. I'm over trying to do this with Github actions.

* https://docs.gitlab.com/ee/ci/ci_cd_for_external_repos/github_integration.html
* https://about.gitlab.com/solutions/github/

Shelob9:
OK, so we now have Github actions triggering running PHP unit tests and JavaScript unit tests. Gitlab CI can run WordPress tests.

I set Gitlab up to mirroring it, which should work -> https://gitlab.com/caldera-labs/Caldera-Forms It doesn't, but I suspect that is beacuse it's not in master branch.

I do have tests working in my personal Gitlab fork, beacuse gitlab, where I was working directly in master branch: https://gitlab.com/Shelob9/Caldera-Forms/-/jobs/263367658

@New0 Look at the testing errors in that run. I don't get that when running these locally, but that's something we should address.

New0:
@Shelob9 do you want to open an issue to address those unit tests errors ? I had not these error locally either.

Shelob9:
The docker container we run tests in locally is using phpunit 6 https://github.com/chriszarate/wordpress-phpunit/blob/master/Dockerfile#L1

Shelob9:
Related to the errors we're seeing here:
https://gitlab.com/caldera-labs/Caldera-Forms/-/jobs/264237774

This SO question: https://stackoverflow.com/questions/43762717/phpunit-6-1-x-throws-array-merge-error-when-my-test-class-uses-its-own-constru

Made me look and this test constructor https://github.com/CalderaWP/Caldera-Forms/blob/master/tests/test-run-calculation.php#L242

@New0 Turns out extending the class like that's an issue. I'm going to move setup to setUp method, which is best practice. FYI for next time.

Shelob9:
@New0 I think this is good now. I just need to somehow get Travis to stop blocking my PRs. Apparently deleting Travis file, disabling repo in Travis and removing Travis + Github integration is not how you do that.

PR #3318, opened 2019-08-09
Finish trim only at the end of calculation

### LINKED ISSUE #3316: round Function is causing a blank total in the new update (manual format)
round Function is causing a blank total in the new update  manual format)
caldera version 1.8.6 
php version 7.3.7
WP Version 5.2.2
it was working fine before but now it just gives me a blank result

New0:
Fixes #3316 

Only finish trim at the end of string when manual calculation is saved.

We decided in last update that we would trim manual calculations in case someone copy and paste them, I introduced this bug that removed the r for round() function doing so. 

To manually test this :
 - insert a calculation field in the form settings and set the calculation to be manual. Then use round to set something like `round(2 * 2.3333)`.
- In the current develop branch, when you save the form and refresh the page the formula will be transformed to `ound(2 * 2.3333)`
- If you switch to feature/3316 , set the  `round(2 * 2.3333)` again, save the form and refresh the page, the formulay should still be  `round(2 * 2.3333)`

I also added a unit test, to run them you will need to install the dev environment using composer dev:install and then run composer test:php


tests/test-forms-api.php
```diff
@@ -235,5 +235,35 @@ public function testManualFormulasTrimDuringSave ()
...
+	 * @since 1.8.6
+	 *
+	 * @covers Caldera_Forms_Forms::save_form()
+	 */
+	public function testManualFormulasTrimDuringSaveWithRound ()
+	{
+		$form_id = $this->import_contact_form();
+		$form = Caldera_Forms_Forms::get_form($form_id);
+
+		$form['fields']['fld_60011111'] = [
```
mrpritchett: I know GitHub can make alignment all funky, but I want to make sure things are aligned properly even if GitHub doesn't show it properly. :)

tests/test-forms-api.php
```diff
@@ -235,5 +235,35 @@ public function testManualFormulasTrimDuringSave ()
...
+		];
+		Caldera_Forms_Forms::save_form($form);
+
+		$form = Caldera_Forms_Forms::get_form($form_id);
+
+		$this->assertSame( 'round(2 * 2.3333)', $form['fields']['fld_60011111']['config']['manual_formula'] );
+	}
+
 
 }
```
mrpritchett: Newline at end of file.

mrpritchett:
This is good to go on my end.

Issue #3345, opened 2019-09-11
Programatically created Form does not play well with DB created Form


Luc45:
Caldera relies on the `caldera_forms_get_forms` filter to provide an "existence" to the Forms that were created programmatically and does not exist in the database.

That filter is applied in `\Caldera_Forms_Forms::get_forms`: https://github.com/CalderaWP/Caldera-Forms/blob/master/classes/forms.php#L178-L199

It applies the filter at line 186, but overrides it again to the original value on line 198.

Despite that, methods such as `\Caldera_Forms::render_form` might continue working, because of a `!empty($forms)` assertion. Check line 4278 of `core.php`: https://github.com/CalderaWP/Caldera-Forms/blob/master/classes/core.php#L4278

That probably means that two bugs co-exist generating an erratic behavior. If `$forms` return empty, it will not try to check for the existence of the Form, thus will continue the processing and will actually output the form. If `$forms` contain anything (such as a form saved on the database), it will assert if the form ID we are trying to render exists. That will fail because the filter was overridden, so `render_form` will just return null.

Possible solution:

Before:
```
if( $with_details ){
	$forms = self::add_details( static::$index );
}else{
	$forms = static::$index;
}
```

After:
```
if( $with_details ){
	$forms = self::add_details( static::$index );
}
```

Luc45:
PR opened: https://github.com/CalderaWP/Caldera-Forms/pull/3349

Shelob9:
@Luc45 Thanks for the detailed report. I opened #3354 which SHOULD fix this.

ictbeheer:
Could #3354 please be merged? It is working for me, but now I have to manually apply this patch after every CF update to be able to use external forms.

Shelob9:
@New0 is this one on your list for next update? Should be easy win.

New0:
Closed via https://github.com/CalderaWP/Caldera-Forms/commit/39f19c2829c80e45035be0deb4d11f66047e0611

Issue #3341, opened 2019-09-08
External forms are not available in the widget


ictbeheer:
## What Version Of Caldera Forms, WordPress and PHP Are You Using?
WordPress Version: 5.2.3PHP Version: 7.2.22MySQL Version: 5.6.45Caldera Forms Version: 1.8.7WP_DEBUG: 1

## Does Your Issue Persist When You Disable All Other Plugins and Switch To The Default Theme?
Yes

## What Is The Unexpected Behaviour?
When external forms are included using the 'caldera_forms_get_forms' they are not available in the widget.

## Cause
This happens because the variable $forms in which they are stored gets overwritten a few lines later.

https://github.com/CalderaWP/Caldera-Forms/blob/ff614a4bd16381f92d6314a0d2f907f4951938d2/classes/forms.php#L186-L199

## What PHP Errors Have You Logged While Reproducing This Bug?
none

## What JavaScript Errors Have You Seen While Reproducing This Bug?
none

ictbeheer:
Is same as #3345 which has a pull request.

Shelob9:
I'm reopening this as this describes an effect of the low-level bug described in #3349

Shelob9:
SHOULD be fixed by #3354

New0:
Closed via https://github.com/CalderaWP/Caldera-Forms/commit/39f19c2829c80e45035be0deb4d11f66047e0611

PR #3349, opened 2019-09-13
Fix programmatically created Form not working when there's a Form created through the UI

Luc45:
Caldera relies on the `caldera_forms_get_forms` filter to provide an "existence" to the Forms that were created programmatically and does not exist in the database.

That filter is applied in `\Caldera_Forms_Forms::get_forms`: https://github.com/CalderaWP/Caldera-Forms/blob/master/classes/forms.php#L178-L199

It applies the filter at line 186, but overrides it again to the original value on line 198.

Despite that, methods such as `\Caldera_Forms::render_form` might continue working, because of a `!empty($forms)` assertion. Check line 4278 of `core.php`: https://github.com/CalderaWP/Caldera-Forms/blob/master/classes/core.php#L4278

That probably means that two bugs co-exist generating an erratic behavior. If `$forms` return empty, it will not try to check for the existence of the Form, thus will continue the processing and will actually output the form. If `$forms` contain anything (such as a form saved on the database), it will assert if the form ID we are trying to render exists. That will fail because the filter was overridden, so `render_form` will just return null.

This PR fixes the behavior that is overriding the filter `caldera_forms_get_forms`.

Original issue: https://github.com/CalderaWP/Caldera-Forms/issues/3345

### DIFF
--- a/classes/forms.php
+++ b/classes/forms.php
@@ -175,6 +175,8 @@ class Caldera_Forms_Forms {
 		    static::$index = static::get_stored_forms();
         }
 
+        $forms = static::$index;
+
         if ( false === $internal_only ) {
             /**
              * Runs after getting internal forms, use to add forms defined in file system
@@ -183,7 +185,7 @@ class Caldera_Forms_Forms {
              *
              * @param array $base_forms Forms saved in DB
              */
-            $forms = apply_filters('caldera_forms_get_forms', static::$index);
+            $forms = apply_filters('caldera_forms_get_forms', $forms);
             if (!empty($forms) && is_array($forms)) {
                 foreach ($forms as $form_id => $form) {
                     $forms[$form_id] = $form_id;
@@ -193,10 +195,8 @@ class Caldera_Forms_Forms {
 
 
 		if( $with_details ){
-			$forms = self::add_details( static::$index );
-		}else{
-			$forms = static::$index;
-        }
+			$forms = self::add_details( $forms );
+		}
 
 		if( $orderby && ! empty( $forms ) ){
 			$forms = self::order_forms( $forms, $orderby );


kjohnson:
@New0 @Shelob9 could you take a look over this, also? The changes make sense to me, but I'm still new to the ecosystem.

Shelob9:
@kjohnson I'll review. It **looks** right. I don't like that there is no test coverage for the bug, this is an API we don't want to break. I'll add a test or two.

Shelob9:
I changed the base branch to develop it is still clean.

Shelob9:
I am testing this branch with two forms saved in database and one file based. The file based field does NOT show up as an option in the Gutenberg block.

If I add the file-based form to the page with PHP, it loads the right form:

```php
add_filter( 'the_content', function(){
    return Caldera_Forms::render_form('file_based' );
});
```

Shelob9:
I would EXPECT file-system based forms to be selectable in the block, both beacuse that makes sense and we call `Caldera_Forms_Forms::get_forms()` without the second arg, which defaults to false for internal forms only.

https://github.com/CalderaWP/Caldera-Forms/blob/e3ca75aa2975213c796937b1a8249a70f37c1188/ui/blocks/init.php#L41

Shelob9:
Same behavior with this branch and the develop branch. Form doesn't show in widget or Gutenberg block. So #3345 and #3341 are real bugs. I'm going to push a test to prove it.

Shelob9:
@Luc45 Thank you for the pull request. This helped a lot towards getting to the right solution #3354 

Thank you again for your contribution. PRs are always welcome. We maintain [a contributor's guide](https://github.com/CalderaWP/Caldera-Forms/blob/master/CONTRIBUTING.md) feel free to take a look or ask questions.

Closing in favor of PR #3354

Issue #1683, opened 2017-06-18
Color picker field shows color code only


101010f:
version 1.5.1

Is this how the color picker should render? I remember having a color preview on one of the sides beside the color hex code.

Using default theme and disabled all plugins.

![sketch](https://user-images.githubusercontent.com/20680030/27260262-79209b62-5430-11e7-8661-4f8c55527c24.png)

Shelob9:
Yes, that is how it did work in CF 1.4

Shelob9:
Funny thing is that the span to show it is there, and the CSS that should make it shown is there... We did change in 1.5.1 how it gets instantiated. Before [no settings](http://labs.abeautifulsite.net/jquery-minicolors/#settings) were passed, now all settings are passed as default and there is a filter, which is cool, but like would be cooler if it worked.

101010f:
The color picker seems to be dead for good with the latest update 1.6.3 Only shows a text input field

hellohamzak:
is there any update on Color picker i find is not really working at all. on last couple of release with using only  twenty seventeen theme & plugin

kjohnson:
@New0 `color_picker_init()` is targeting `.minicolor-picker`, which looks to be the class on the wrapper element. Instead, we should be targeting the `<input />` element directly.

New0:
Closed via https://github.com/CalderaWP/Caldera-Forms/commit/f68e48f49d06f502fd93012a73b69691f3f22eb4

PR #3353, opened 2019-09-19
Target correct file to generate <input> for Color Picker field

New0:
For #1683 

The color picker configuration used the generic-input.php file to generate the <input> for the color picker field when the color picker input works with color_picker/field.php

### DIFF
--- a/classes/field/input.php
+++ b/classes/field/input.php
@@ -62,7 +62,11 @@ class Caldera_Forms_Field_Input extends Caldera_Forms_Field_HTML{
 			$place_holder  = self::place_holder_string( $field, $field[ 'label' ] );
 		}
 
-		if( 'number' === $type ){
+		if('color_picker' === $type ){
+			$attrs[ 'class' ] = 'form-control minicolor-picker init_field_type miniColors';
+			$attrs[ 'type' ] = 'text';
+			$attrs[ 'data-type' ] = 'color_picker';
+		}elseif( 'number' === $type ){
 			foreach( array(
 				'min',
 				'max',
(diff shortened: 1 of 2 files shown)


New0:
@Shelob9 I have second thoughts about this, was it intentional to target generic-input.php ? Maybe with the idea to set the input attributes in classes/field/input.php

kjohnson:
@New0 looks like @Shelob9 did intentionally change to a generic input, see https://github.com/CalderaWP/Caldera-Forms/commit/254f60eab17c1f03d9ee7ab966ed8d9a884f399e, although this may have been a regression.

Shelob9:
@New0 Right. I wanted to move all input fields to generic input, so that the input's attributes would pass through `caldera_forms_field_attributes` filter.

If @kjohnson is correct here about the class name being the issue - https://github.com/CalderaWP/Caldera-Forms/issues/1683#issuecomment-532763280 - then that filter could be used to fix it. Or maybe this indicates that  `Caldera_Forms_Field_Util::prepare_field_classes()` is defective.

Also, take a look at the `caldera_forms_field_js_color_picker_settings` filter.

kjohnson:
The change to the generic would explain why the render didn't match what I was seeing in the codebase. It could be as simple as the generic isn't setting the correct class attribute - which would have been an understandable oversight without an acceptance test.

New0:
Thank @Shelob9 , yes @kjohnson , it seems like the precise attributes required by the color picker field were not set during the file switch.
It runs through caldera_forms_field_attributes filter hook.
I will add a Ghost Inspector test for the color picker and report before we merge.

New0:
Created test https://app.ghostinspector.com/tests/5def652741fe8157e17697ef

PR #3354, opened 2019-09-19
Ensure `\Caldera_Forms_Forms::get_forms:()` returns forms added on a filter.

### LINKED ISSUE #3341: External forms are not available in the widget

### LINKED ISSUE #3345: Programatically created Form does not play well with DB created Form

Shelob9:
This PR fixes #3345 

It also fixes #3341

When I test with this branch, I see the form added using a filter int he Gutenberg block, form entry viewer and widget.

I think I got both of the bugs that @Luc45 said ` co-exist generating an erratic behavior` [here](https://github.com/CalderaWP/Caldera-Forms/issues/3345#issue-492514529).

I prevented  the overwrite of `$forms` here:
https://github.com/CalderaWP/Caldera-Forms/pull/3354/files#diff-634fa6588d7d78d435ea1ea46e18563cR195-R199

I also prevented forms added on a filter from being removed here:
https://github.com/CalderaWP/Caldera-Forms/pull/3354/files#diff-634fa6588d7d78d435ea1ea46e18563cL417

Note, I am intentionally bypassing reading from caching of that variable. I'm not worried about the performance, I am worried about bugs caused by WHEN the filter runs.

Also, I expanded the test that SHOULD have been covering this filter, but didn't tests these situations. I left my assumptions as inline comments:
https://github.com/CalderaWP/Caldera-Forms/pull/3354/files#diff-8c0dd687d0ffb15b67072a3555cb2700L73

### DIFF
--- a/classes/forms.php
+++ b/classes/forms.php
@@ -189,14 +189,14 @@ class Caldera_Forms_Forms {
                     $forms[$form_id] = $form_id;
                 }
             }
+
         }
 
+		$forms = isset($forms) && false === $internal_only ? $forms : static::$index;
 
 		if( $with_details ){
-			$forms = self::add_details( static::$index );
-		}else{
-			$forms = static::$index;
-        }
+			$forms = self::add_details( $forms );
+		}
 
 		if( $orderby && ! empty( $forms ) ){
 			$forms = self::order_forms( $forms, $orderby );
@@ -411,15 +411,11 @@ class Caldera_Forms_Forms {
 	 */
 	protected static function add_details( $forms ){
 	    if( empty( $forms ) ){
-	        return [];
+			return [];
         }
 
-        if( ! empty( $valid_forms = get_transient( self::$registry_cache_key ) ) ) {
-            return $valid_forms;
-        }else{
-            $valid_forms = [];
-        }
-		
+		$valid_forms = [];
+
 		foreach( $forms as $id => $form  ){
 			$_form = self::get_form( $id );
 			if( empty( $_form ) ){
@@ -442,9 +438,6 @@ class Caldera_Forms_Forms {
 				}else {
 					$valid_forms[ $id ][ $key ] = '';
 				}
-
-
-
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


classes/forms.php
```diff
@@ -411,15 +411,11 @@ protected function force_string( $val ){
...
+			return [];
         }
 
-        if( ! empty( $valid_forms = get_transient( self::$registry_cache_key ) ) ) {
-            return $valid_forms;
-        }else{
-            $valid_forms = [];
-        }
-		
+		$valid_forms = [];
```
kjohnson: In the PR you callout intentionally bypassing the cache, but once merged that decision will be buried, so it may be helpful to add a comment to explain the intent of bypassing the cache.

Shelob9: Good point. Added in https://github.com/CalderaWP/Caldera-Forms/commit/1a6a1e0c69fa2152df0e64e5ac8dc2c8e937ba88

Shelob9:
@New0 Your change in 700abcc4ac6a436de34a5f40aaa32f356871df56 makes sense to me. Either I counted wrong or we added another form there since I wrote this. That's why this type of test is NOT ideal. But, it's fine for now.

Issue #3356, opened 2019-09-20
PHP Notice:  Undefined variable: symbols in /classes/core.php on line 1416


New0:
It looks like the order of the condition is causing that notice l1416

New0:
Closed via https://github.com/CalderaWP/Caldera-Forms/commit/9dd2900dc11a2f3aa41c42c29bc72ea588505250

PR #3357, opened 2019-09-20
Swap order of conditions to prevent PHP undefined index notice

New0:
For #3356 

The condition was checking the type before checking if not empty, swapping the order should prevent the notice.


New0:
@kjohnson Could you review that PR?

Issue #3358, opened 2019-09-20
Map wp-content/plugins in docker-compose so it's easier to work with add-on plugins


Shelob9:
I always do this locally, meant to commit it. If we map wp-content/plugins in docker-compose, adding an add-on or in my case Ninja Forms, adding an additional plugin is easy, just add it to the directory.

New0:
Closed via https://github.com/CalderaWP/Caldera-Forms/pull/3359

PR #3359, opened 2019-09-20
Map wp-content/plugins into docker container 

Shelob9:
Simple PR that changes the path mappings in docker-compose

To test, copy a plugin into `wp-content/plugins` and see if it works. Now I can install Ninja Forms!

#3358

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,6 +1,7 @@
-version: "3.1"
+version: '3.1'
 
 services:
+
   wordpress:
     image: wordpress
     ports:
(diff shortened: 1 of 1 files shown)


Shelob9:
>It works as per the explanation, I also tested uploads to media library from file fields as the uploads path map was removed

@New0 Did it work? I didn't think about that.

New0:
> 
> 
> > It works as per the explanation, I also tested uploads to media library from file fields as the uploads path map was removed
> 
> @New0 Did it work? I didn't think about that.

Yes it worked, I xould have files uploaded to the media library using WordPress admin and also using the upload to media library option of File fields

Shelob9:
@New0 That's great.

PR #3367, opened 2019-09-30
add `Release Asset` header

afragen:
Make compatible with GitHub Updater and  Caldera-Forms build process.


afragen:
Without this header anyone updating from GitHub Updater will have an un-built plugin. 🥺

afragen:
Hey Josh! I miss seeing you too. I hope we can all get back to normal soon. Be well. 

Essentially this is just a header comment that only interacts with GitHub Updater.

PR #3414, opened 2019-12-05
Hide CF pro tab and menu if not active or filter is not overiding

### LINKED ISSUE #3413: Hide CF Pro menu/ UI if CF Pro not active
To reduce confusion, only show CF Pro menu/ tab IF it is active.

Add a filter that sets CF Pro UI to show or not. 
```
/**
 * Should we add Caldera Forms Pro sub-menu page and UI
 *
 * @since 1.8.9
 *
 * @param bool $show_pro Whether to show or not
 */
$show_pro = apply_filters( 'caldera_forms_show_pro_ui', caldera_forms_pro_is_active() );
```

If Caldera Forms Pro is activated, that filter would default to true, and UI would show.  If Caldera Forms Pro is activated, that filter would default to true, and UI would show. 

If Caldera Forms Pro is not active to activate it, customer would have to add this filter:

```
add_filter( 'caldera_forms_show_pro_ui', '__return_true' );
```

Shelob9:
Fixes #3413

### DIFF
--- a/classes/admin.php
+++ b/classes/admin.php
@@ -1768,16 +1768,6 @@ class Caldera_Forms_Admin {
 							'text' => __( 'Processors getting started guide', 'caldera-forms' )
 						)
 					),
-                    "antispam" => array(
-                        "name" => __( 'Anti-Spam', 'caldera-forms' ),
-                        "location" => "lower",
-                        "label" => __( 'Anti Spam', 'caldera-forms' ),
-                        "canvas" => $path . "anti-spam.php",
-                        'tip' => array(
-                            'link' => 'https://calderaforms.com/doc/protect-form-spam-caldera-forms/?utm_source=wp-admin&utm_medium=form-editor&utm_term=tabs',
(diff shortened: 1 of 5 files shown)


Shelob9:
@New0 I think, to reduce the page load size, we should prevent Vue from ever loading in edit screen. As long as there is no way to add back the "Pro" tab. I think this PR should:

- Delete the "Pro" tab.
- Hide the pro submenu item unless CF Pro is active.
- Have a filter to force the CF Pro submenu back in, so that the submenu can be used.

New0:
@Shelob9 OK, we cross commented... So I understand we still need to completely remove the Pro tab from the form edition page and load Vuejs only on the Pro page. That page will be loaded if the user is connected to Pro or using the filter 'caldera_forms_show_pro_ui' to restore it.

Is that correct?

Shelob9:
> Is that correct?
Yes.

>Also shouldn't we remove the anti-spam tab using the same method? And restoring the Honeypot option in te form settings?
Yes.

New0:
@Shelob9 could you review the PR?

PR #3417, opened 2019-12-10
Prevent get_field_data() from always returning null for string form IDs

telwing:
The `$form` variable containing the original form ID is being overwritten too soon. The condition `$form['ID'] !== $form` will always return true because `$form['ID']` is being compared to the `$form` object.


New0:
Thank you @telwing !

Issue #1354, opened 2017-03-05
Refactor conditionals in form editor


Shelob9:
Goals:
- Improved editor load time
- Don't store config in element attribute, use form state representation created in #1139 instead
- Setup ability to have conditionals be in multiple fields
- Fix #247

Issue #3437, opened 2020-01-08
Run php and JS unit tests in gitlab ci


Shelob9:
The Github actions we have do not work -- old format. Gitlab is already running WordPress integration tests.

123aqua:
@Shelob9 Hi can you help me with something When I fill up phone number it says undefined I'm using caldera forms here is the screen shot
![16](https://user-images.githubusercontent.com/59171353/72647626-ead55480-39b3-11ea-987d-fc17df279a99.png)

Shelob9:
Close via #3438

Shelob9:
@123aqua Thanks for using Caldera Forms, I appreciate it. I've deleted your comment, as it is not related to this issue. If you belive you have found a bug, please open a new issue and fill out all of the requested information. If you want to submit a fix for the bug, that's super cool, please [review our contributor's guide before submitting a PR](https://github.com/CalderaWP/Caldera-Forms/blob/master/contributing/pull-request-workflow.md)

123aqua:
Hi can you help me with something When I fill up phone number it says
undefined I'm using caldera forms here is the screen shot

Issue #3435, opened 2020-01-07
New state mananagment API in form editor to fix performance


Shelob9:
Blocked by: #3240

The form editor is SLOW when there are many fields and esepcially so when there are many conditionals. Root problem:

- Every time a field or conditional changes its settings, we rebuild the magic tag list from scratch.
- The state of fields and conditional is read and written to the DOM on every update to a field or conditional.

See: #1138 #1137 #1136 #1139 #1354

I am proposing to introduce a new API that is:

- The one true source of state for conditionals -- storing there settings in the DOM is the biggest problem and probably the easiest one to replace.
- Knows enough about fields and system tags to generate an up to date magic tags auto-selector. Fields and processors will -- for now -- still write there data to the DOM.

![cf-editor-state](https://user-images.githubusercontent.com/1994311/71904737-b2708380-3134-11ea-8e53-6db8a19810fa.jpg)

Shelob9:
That diagram is from yesterday. Looking at it and the protoype now, I see that it has two big problems:

_ No way to get all tags of a type. For example, all email fields.
- It's unware of what the tags get wrapped in - `%tag%` vs `{tag}`
- Processor's add tags that are not incluced in this schema.


```js
{
	"field": {
		"tags": {
			"text": [
				"header",
				"first_name",
				"last_name",
				"email_address",
				"message",
				"comments_questions",
				"submit"
			],
			"html": ["header", "message"],
			"email": ["email_address"],
			"paragraph": ["comments_questions"],
			"button": ["submit"]
		},
		"type": "Fields",
		"wrap": ["%", "%"]
	},
	"system": {
		"type": "System Tags",
		"tags": {
			"text": [
				"entry_id",
				"entry_token",
				"ip",
				"user:id",
				"user:user_login",
				"user:first_name",
				"user:last_name",
				"user:user_email",
				"get:*",
				"post:*",
				"request:*",
				"post_meta:*",
				"embed_post:ID",
				"embed_post:post_title",
				"embed_post:permalink",
				"embed_post:post_date",
				"date:Y-m-d H:i:s",
				"date:Y/m/d",
				"date:Y/d/m",
				"login_url",
				"logout_url",
				"register_url",
				"lostpassword_url",
				"referer_url"
			],
			"email": ["user:user_email"],
			"date_picker": ["embed_post:post_date", "date:Y-m-d H:i:s"]
		},
		"wrap": ["{", "}"]
	},
	"increment_capture": {
[...]

PR #3442, opened 2020-01-13
Add cypress.io tests for conditional logic editor

Shelob9:
This PR adds tests for the conditonal logic editor so they can get there performance issues safely fixed via #1354


Shelob9:
This branch currently includes #3437 and #3435

Shelob9:
The "Add Conditional Line" button adds a group of lines. The "Add Condition" button adds a line to a group of lines. The labels of those buttons disagree, but these are the true facts.

lines-vs-group

Shelob9:
7ed5741 cb6eadb 70a7240 in PR #3442 Add cypress.io test coverage for creating a form and a lot of conditonal logic. I am ALMOST comfortable with removing the current state management for conditional logic and then rewiring it to new system. Coverage I still want

- [ ] Switching in and out of conditonal editor -- basically done.

- [ ] Coverage for multiple conditional groups


- [ ] The options list drop down is use if the conditional is based on a select/radio/checkox/etc field.

- [ ] Given conditional is based on a radio field, assert has  options list drop down for conditional. Given field type has changed to type, assert that the conditional can now be set with  text 

- [ ] Given conditional is based on a radio field, assert has  options list drop down for conditional. Given that an additional option has been added for that radio field, assert that options list drop down for conditional has new option.

- [ ] Saved conditionals load properly. Should test 1) given we've loaded one of the forms we already have that use that has conditionals assert, conditional editor loads with the right state. 2) given that we created a form and added conditionals, save form and  reload page, assert conditional editor loads with the right.

Shelob9:
@New0 This PR is now only for adding additional tests. I will branch off of this branch to impiment the changes via a new PR. That way, even if that ends up not working out, we have these tests.

This branch includes #3437 and #3435 and should get merged after those ones.

PR #3438, opened 2020-01-08
run js unit tests in gitlab ci

Shelob9:
#3437

Issue #1138, opened 2017-01-20
edit.js build_field_preview is, like, SUPER slow


Shelob9:
`data_fields.each(function(k,v){` is offender

Issue #1137, opened 2017-01-20
edit.js rebind_field_bindings is a SUPER slow function


Shelob9:
Profiled using 1.4.8, and real big form, took 764ms,

- loop `for(var v = 0; v < bindings.length; v++){` 24ms
- It interacts with jQuery UI Draggable, which is a huge consumer of resources according to this profile.

Shelob9:
DELETED! Close via 	0f37381

Issue #1136, opened 2017-01-20
edit.js init_magic_tags does a SUPER slow loop magicfields.each(function(k,v){


Shelob9:

Shelob9:
DELETED! Close via 	0f37381

PR #3445, opened 2020-01-14
Improve performance of form editor with lots of conditionals

### LINKED ISSUE #1354: Refactor conditionals in form editor

Shelob9:
This PR will add a new state managment API #3435 and use it to manage the state of conditonal logic .

fixes #1354

This is first step to fixing performance issues such as #1138 and #1137 It will NOT address the magic tags list being slow #1136. Once this is merged we can either decide that wasn't such a big deal or address that issue using this same state managment API next.

Blocked by:

#3442 and #3438


Shelob9:
At 1dd1baa I think we have unit test coverage for new API. The commits included from #3442 have e2e test coverage to prove that the conditional logic editor works. I am now prepared to rewire conditional logic's state managment.

Shelob9:
Components added and slightly implimented in 9b5628af25fea09490ae982d2e1148e10e589886 will replace condtional logic with more performant system.

It is WIP, I feel good about getting done. The tests in  #3442 will prove this works, when it does.

Need to add:
- [ ] fully wire new api
- [ ] Layout componets correctly
- [ ] Figure out processor conditionals
- [ ] Snapshots test for the main components. 
- [ ] Maybe unit tests for these components' state, but I think that might be over kill.

Shelob9:
What this has:

* Completely rewritten conditonals tab in React. It does everything the old one did, but WAY faster. I added 40 conditonals and it didn't affect performance at all.

What is left to do:

- [ ] When fields update, update there values in `window.cfEditorState`
- [ ] Make conditional editor aware of field options for select/checkbox/etc fields.
- [ ] The checkboxes for what fields to apply to disable/enable state updates one render later than it should.
- [ ] Save conditionals (keeping in mind that I started storing which fields it conditonals apply to with the conditonal, not the field, but probably need to have it in both places, see 095e04c8c4f05adbacf5f69912d01a24c06134b5 )
- [ ] [Setting intial conditonal state in PHP ](https://github.com/CalderaWP/Caldera-Forms/blob/feature/3425-2/classes/admin/assets.php#L358-L399) needs test coverage.
- [ ] Conditional logic for processors.

Shelob9:
Closing in favor of #3519

PR #3461, opened 2020-01-22
Allow processor template to be a string 

### LINKED ISSUE #3460: Allow processor template to be a string
Currently the processor UI template is loaded with include. https://github.com/CalderaWP/Caldera-Forms/blob/master/ui/panels/processors.php#L315-L320

For the email-crm project, I need to be able to generate the UI as a string of HTML.

Shelob9:
Fixes #3460

Currently, if `$config['template']` isset, that file is included. This PR maintains that as the default. But, it also adds a fallback to see if `$config['html']` isset, if so, that key of the array is echoed.

@New0 I need this feature in next release in order to do Mailchimp and Salesforce add-ons.


ui/panels/processors.php
```diff
@@ -315,7 +315,11 @@ function build_processor_types($default = null){
 	if(isset($config['template'])){
 		include $config['template'];
 	}else{
-		echo '<p>' . __('This processor has no configurable options.', 'caldera-forms') . '</p>';
+		if( isset($config['html'])){
```
kjohnson: Instead of a nested `if`, should we instead use flatten the check into an `ifelse`?

Shelob9: @kjohnson Yes, good point. I updated.

PR #3464, opened 2020-01-22
Add forms and processor data services to cf2 container

### LINKED ISSUE #3463: Introduce new cf2 services to aid in add-on development
Related: https://git.saturdaydrive.io/_/engineering-ops/sd-email-crm/email-crm/issues/20 and #3134

## Problem

When developing the add-ons in email-crm project, we need to be able to test processors isolated from Caldera Forms. To do so, we need:

* An interface describing forms API that this also implimented in Caldera Forms core, that we can replace with a mock. I can't mock out `Caldera_Forms_Forms` methods beacuse you can't mock a static class method.

* An interface describing transients API that this also implimented in Caldera Forms core, that we can replace with a mock. [Exists already](https://github.com/CalderaWP/Caldera-Forms/tree/master/cf2/Transients)
* A factory for `Caldera_Forms_Processor_Get_Data` objects -- this class has internal dependencies to hard to mock methods of `Caldera_Forms` class.
* A way to get the real functional version of these services into the add-ons. [Can be done with existing container](https://github.com/CalderaWP/Caldera-Forms/blob/master/cf2/CalderaFormsV2.php) by adding new services

## Solution

Instead of creating this in the email-crm project, and maintaining it there, even though it would only break if core breaks, I will add new services to the CF2 container.

To ensure these service work, I will add unit and integration tests.

Shelob9:
Resolves #3463 by adding and binding to the cf2 container:

- [x] A forms service that can get and update forms.
- [x] A factory for Caldera_Forms_Processor_Get_Data objects

### DIFF
--- a/classes/forms.php
+++ b/classes/forms.php
@@ -636,8 +636,10 @@ class Caldera_Forms_Forms {
 		$added = self::save_to_db( $newform, 'primary' );
 		if( ! $added ){
 			return false;
-
 		}
+
+		// Fixes https://github.com/CalderaWP/Caldera-Forms/issues/3455
+		self::clear_cache();
 
 		/**
 		 * Runs after form is created
--- a/cf2/CalderaFormsV2.php
+++ b/cf2/CalderaFormsV2.php
@@ -7,172 +7,177 @@ namespace calderawp\calderaforms\cf2;
 use calderawp\calderaforms\cf2\Fields\FieldTypeFactory;
 use calderawp\calderaforms\cf2\Transients\Cf1TransientsApi;
 use calderawp\calderaforms\cf2\Services\ServiceContract;
+
 class CalderaFormsV2 extends \calderawp\CalderaContainers\Service\Container implements CalderaFormsV2Contract
 {
 
-	/**
-	 * Path to main plugin file
-	 *
-	 * @since 1.8.0
-	 *
-	 * @var string
-	 */
-	protected $coreDirPath;
+    /**
+     * Path to main plugin file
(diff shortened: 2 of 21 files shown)


Shelob9:
Using this branch in email-crm project, I have a test showing that I can resolve mock form service and processor factory here: https://git.saturdaydrive.io/_/engineering-ops/sd-email-crm/email-crm/commit/f5a135e19c1dcc9781fcb30af0f6a2cd74abe267#208b2c3cf154ca7c5a50ef1a43316d524cac0a24_21_38

Shelob9:
Built at 8fb4ac668063944d9d2ede033b3f37166107baaa [cf-3463.zip](https://github.com/CalderaWP/Caldera-Forms/files/4104031/cf-3463.zip)

classes/forms.php
```diff
@@ -636,9 +636,11 @@ public static function create_form( $newform ){
 		$added = self::save_to_db( $newform, 'primary' );
 		if( ! $added ){
 			return false;
-
 		}
 
+		// Fixes https://github.com/CalderaWP/Caldera-Forms/issues/3455
+		self::clear_cache();
```
kjohnson: Is this change relevant to the PR, or is this change a stowaway?

Shelob9: It is a stoway from #3456 therefore this should NOT get merged before #3456 

I need to be more careful.

kjohnson: #3456 is showing as merged - resolving this conversation.

Issue #3374, opened 2019-10-14
Variables set in processors are unset after form edition


New0:
Reported via - https://secure.helpscout.net/conversation/950098119/23002/ -

After creating a variable and setting the variable in a processor, the variable is unset from the processor whenever the form is edited.

If the form is not edited(re-saved) the variable could be not showing but is still set during form processes and appears to work.

New0:
Seems also related - https://secure.helpscout.net/conversation/980498333/0/?folderId=832038 -

lukebullock:
Can confirm this is happening on a site I manage. 

It is extremely frustrating when you have more than 30 processors. The variables are removed from the processor and unavailable from the dropdown selector even though the variables still show up under the variables tab. If you create a new variable the old variable will be made available in the dropdown selector in the processor again.

**The way I am having to workaround this bug:** 

1. Create a new unnecessary variable and then return to the conditional processor; the old variable should now show up in the dropdown list and be selectable...select it.
2. Delete the unnecessary variable you just created.
3. Rinse and repeat for every single processor using the variable. 
4. Save form.
5. Pray that client doesn't want any more updates.

Shelob9:
Close via #3466

PR #3466, opened 2020-01-23
Add variables to magic tags during page load

New0:
For #3374 It appears variables are added to the list of system tags that fill the list of magic tags available in fields during the creation of variables only. If the page is reloaded the variables are not listed anymore.

The function applied on change event of the field to create a variable to add variables to system tags for the magic tags list is also applied on page load in this PR.

- [x] Create a test

- [x] Apply the function to add variables to systme tags on page load


cypress/support/util.js
```diff
@@ -502,4 +502,24 @@ export const createForm = (name,blankForm = true) => {
...
+
+/**
+ * Variables functions
+ * 
+ * @since 1.8.10
+ */
+export const cfGoToVariablesTab = () => {
+	cy.get('#tab_variables a').click();
+};
+export const cfAddVariable = () =>{
```
Shelob9: @New0 missing inline docs. It was really helpful to me when I looked at these functions for first time in a year, that they had one line describing what they do.

New0: You're right I corrected that. I had it set like a header for a group of paragraphs.

Issue #3403, opened 2019-11-14
Calculation fields do not update when revealed in conditional


Stegosource:
## Do You Need Immediate Help?
It would be nice.

## Is This A Feature Request?
It is not.

## What Version Of Caldera Forms, WordPress and PHP Are You Using?
WordPress Version: 5.3 PHP Version: 7.2.24 MySQL Version: 5.7.28 Caldera Forms Version: 1.8.8 WP_DEBUG: 

## Does Your Issue Persist When You Disable All Other Plugins and Switch To The Default Theme?
Yes

## Steps to reproduce
- Create a form
- Add a number field (no default)
- Add a checkbox field (unchecked by default)
- Add a calculation field that shows the number from the number field
- Create a condition that only shows the calculation if the checkbox is checked.
- Publish and go to form
- Fill in number field
- Check checkbox field 

## Expected behavior
After filling in the number and checking the checkbox, the calculation field should appear and show the calculation based on the number input.

## Actual behavior
After filling in the number and checking the checkbox, the calculation field appears, but the calculation is not implemented. It only fires after the number field is changed again.

## What PHP Errors Have You Logged While Reproducing This Bug?
No PHP errors

## What JavaScript Errors Have You Seen While Reproducing This Bug?
No JS errors.

## Comments
It looks like the conditional fields are not added to the page until the condition is met. Then I assume there is an event listener attached to the calculation fields. If this event listener was also fired on the mounted event, that should solve the issue.

New0:
Closed via https://github.com/CalderaWP/Caldera-Forms/commit/4152f4933216bf69d420d471ee6e8a19d6c20447

PR #3470, opened 2020-01-28
Show calculation value when unhiding the calculation field

New0:
For #3403 

Create a test and run calculation when unhiding the calculation field.


Shelob9:
@New0 OK, I thaught this was a branch to add test coverage. Unaproved then. Good tests.

assets/js/field-config.js
```diff
@@ -707,20 +707,19 @@
 				result = 0;
 			}
 
-			if ( null === lastValue || result !== lastValue ) {
```
New0: This condition was removed and is the only change in the code here.

New0:
@Shelob9 I found that the condition if ( null === lastValue || result !== lastValue ) in field-config.js L710 prevents updating the DOM if last value in State is equal to new value. 
That made the Calculation field not printing the value during the cf.add event as the value is not updated at that moment.
I'm describing the issue as the tests pass but the review needs to think about what consequences that can have.

assets/js/field-config.js
Shelob9: I get why this would cause the bug. I THINK that my intent here was to have this do two things:

- Not run a state mutation when the form is first loaded when lastValue would be null. Unfrtounetly, it could be null beacuse of the conditonal hide. Hence this bug.
- Not run a state mutation for no reason -- `result !== lastValue` . If result and lastValue are equal, then it's not meaningful to update lastValue or mutateState.

Maybe the real issue is that when `lastValue` is null, that could mean one of two things. It could mean field has never been run through this system before or that field is being undhidden. Can we represent one of those states with `undefined` instead? `0` and `false` could be valid field values, so I do not want to use those.

cypress/integration/calculation-triggered-when-unhiding-field.test.js
```diff
@@ -0,0 +1,30 @@
+import {
+    visitPage,
+    cfFieldIsNotVisible,
+	cfFieldIsVisible,
+	cfFieldSetValue,
+	cfFieldCheckValue,
+    cfFieldCheckAllValues
+} from '../support/util';
+
+describe('Calculation value is showed when unhiding the calculation field', () => {
```
Shelob9: @New0 These tests show that the bug is fixed. The comment I left on the change is a concern that is neither proven or unproven. Can you please add a second test case here with a form whose calculation field starts NOT hidden, then is hidden and then unhidden? That's good coverage to have by itself. If those tests pass then ignore my other comment.

cypress/integration/calculation-triggered-when-unhiding-field.test.js
```diff
@@ -0,0 +1,30 @@
+import {
+    visitPage,
```
Shelob9: @New0 Please watch your indents.

Issue #3024, opened 2019-02-27
Editing entires in Number Field with value of 0 returns a blank field


stonge85:
# PLEASE READ THIS LINE & Then Delete - We need this information to best understand your issue, please answer all questions!

## Do You Need Immediate Help?
* If so, please open a support ticket at https://CalderaForms.com/support

## Is This A Feature Request?
* If so, that's fine, please describe your desired featured and delete the rest of this template.

## What Version Of Caldera Forms, WordPress and PHP Are You Using?
*You can copy the "Short Debug" line from the Caldera Forms support menu's debug tab

## What Is The Unexpected Behaviour?
* Please explain what you did, what you expected to happen, and what happened instead.

## What PHP Errors Have You Logged While Reproducing This Bug?
* See: https://calderaforms.com/2016/05/wordpress-debug-logging/

## What JavaScript Errors Have You Seen While Reproducing This Bug?
* See: http://webmasters.stackexchange.com/questions/8525/how-to-open-the-javascript-console-in-different-browsers

stonge85:
When editing a number field entry that is set to 0, the entry form does not populate; meaning the field is blank.  If this is a required field, the entry will not save until the field is repopulated.

stonge85:
This issue has been open for over a year.  How do we get this on the roadmap to get it resolved.

Shelob9:
Added tests in  f692ea7

Bug is most likely introduced here: https://github.com/CalderaWP/Caldera-Forms/commit/f14ec7412a9ca79db0e52c1a10ea535e2ad7ef00#diff-f81f3dff36bf04b7a070d642c858e883

Shelob9:
Here, we're checking if saved value is empty, whioch `'0'` and `0` are.

https://github.com/CalderaWP/Caldera-Forms/blob/master/classes/field/input.php#L21

Shelob9:
Should be fixed via #3471

New0:
Closed via - https://github.com/CalderaWP/Caldera-Forms/commit/d19d7d9f38352468e2f612c9d26f48f34ebbd31d -

PR #3472, opened 2020-01-28
[WIP] Respect conditonal logic for saved entries (fixes resend email issues)

### LINKED ISSUE #3402: Fields applied a condition are not included in resend emails
Reported at -https://secure.helpscout.net/conversation/981366373/23542/ -

When using the resend option in the entry viewer, if fields were applied a condition, the value of the field won't be included in the email. Even in the case the condition is met and a value was given to the field.

Shelob9:
Fixes #3402


Shelob9:
I accidently included #3024 in this.

Issue #3469, opened 2020-01-27
Main mailer is omitting all text between two % symbols in input


New0:
- https://secure.helpscout.net/conversation/1058582555 -

It looks like interference with field magic tags interpretation in the main mailer. When a magic tag failed to be interpreted after reading a % symbol it doesn't replicate the text and just removes it.

New0:
This was already reported by #3157

New0:
Closed via https://github.com/CalderaWP/Caldera-Forms/commit/a59c771c46e6973fabebeb958a315c17e75722eb

PR #3478, opened 2020-01-31
Check that tag retrieved between two % characters in main mailer does …

New0:
…not contain space characters

For #3469 #3157 

That will prevent this content to be interpreted as a magic tag which will not be found and then skipped from content

It needs a unit test to read the content passed to wp_mail


New0:
As the test was added after the commit that fixes the issue, I confirm the test was run with and without the fix. 
The content of the md5 hash used for assertions was analyzed and confirmed it was removing the content between % symbols before the change see  https://github.com/CalderaWP/Caldera-Forms/pull/3478/commits/fb17e7c930b79f722f38e8ec54924c435e52070b

Issue #3484, opened 2020-02-10
Build script does not copy assets files.


Shelob9:
Right here:https://github.com/CalderaWP/Caldera-Forms/blob/develop/Gruntfile.js#L60-L63 We're copying html and CSS, but not the asset.json file that will be needed here: https://github.com/CalderaWP/Caldera-Forms/blob/develop/classes/render/assets.php#L713

This would explain the test failures that @New0 has on the test site, but not in his local.

```
Uncaught TypeError: Cannot read property 'Component' of undefined
    at Object.227 (index.min.js?ver=1.8.10-b1:1)
```

Safe to assume that is `React.Component`. React SHOULD be loaded from `wp-includes` I do not see it in sources. The build process SHOULD have generated the asset.json file that is used to determine what WordPress files to enqueue. We use [this plugin to generate those files](https://developer.wordpress.org/block-editor/packages/packages-dependency-extraction-webpack-plugin/) [here](https://github.com/CalderaWP/Caldera-Forms/blob/develop/webpack.clients.js#L214).

Shelob9:
@New0 Yep, they are not there in your build. Glad you ran those tests.

![Annotation 2020-02-10 120228](https://user-images.githubusercontent.com/1994311/74178609-da5a7780-4c09-11ea-8b4e-9ae39ae87f48.jpg)

New0:
Closed via - https://github.com/CalderaWP/Caldera-Forms/commit/086d21f0bb4e30f0a9d6113ed89f72e2ac9bbf8a -

PR #3485, opened 2020-02-10
Update the list of files to be copied to build for releases

New0:
For #3484 

Copy the assets.json file needed to load react from wp-includes


New0:
@Shelob9 the PHP unit test at tests/integration/Service/FormServiceTest::testGetForms and testGetForm are failing now because it now gets the forms added via filter hook(fixed issue). That makes sense to me but I want to check if you want to work on those tests to clean the forms before counting them again or I can just adapt the count number??

Issue #200, opened 2015-08-19
Front end scripts loading on every page


JiveDig:
Looking on my front page (no form there), and other non-form pages, and the scripts are loading.  I think my suggestion to register the script early, but enqueue it from the shortcode would stop this from happening. See here: https://github.com/Desertsnowman/Caldera-Forms/issues/195

Shelob9:
Probably the same issue as #189

PR #3519, opened 2020-03-20
Replace conditonal logic editors to fix performance issues.

### LINKED ISSUE #3452: Alternative field names, i.e. slugs as field names
* If so, please open a support ticket at https://CalderaForms.com/support

## Is This A Feature Request?
* If so, that's fine, please describe your desired featured and delete the rest of this template.

Hi, I am just starting with Caldera Forms and before buying the Pro version I would like to know if (and how) it is possible to store the input values under slug names (or other self created names) instead of using the field names displayed in the form. 
Example: instead of "Adresse in Hamburg" (displayed in form front end) I would like to use the slug "hh_adresse" or set "hh-adresse" as field name.
The reason is that we are processing the submitted data and would like to keep the field names we used in CF7.
Thanks a lot

Shelob9:
Fixes #3452

This PR adds a new REST API endpoint for saving forms. It also adds a form builder client that uses React for conditional logic, processor conditionals, and the save button. Also, it handles saving to that endpoint.


Shelob9:
@new0 this is now ready for review.

I see tests passing in pipeline.

 Acceptance tests pass locally. I appreciate if you confirm that. Make sure to rebuild clients before running those tests.

cf2/Asset/Register.php
```diff
@@ -0,0 +1,199 @@
...
+     * @since 1.9.0
+     *
+     */
+    public function __construct($handle, $coreUrl, $corePath, array $data = [])
+    {
+        $this->handle = $handle;
+        $this->data = $data;
+        $this->registered = false;
+        $this->setScriptUrl($coreUrl . 'clients/' . $this->handle . '/build/index.min.js');
+        $this->setAssetsFilePath($corePath . '/clients/' . $this->handle . '/build/index.min.asset.json');
```
New0, line 67: From my logs $corePath seems to be set with an ending slash, clients should be set without first slash as on above line ( 'clients/' )

caldera-core.php
```diff
@@ -177,3 +177,24 @@ function caldera_forms_fallback_shortcode()
 	return esc_html__('Form could not be loaded. Contact the site administrator.', 'caldera-forms');
 
 }
+
+
```
New0, line 181: Should we add a dscription here?

cf2/Asset/Register.php
```diff
@@ -0,0 +1,199 @@
...
+
+    /**
+     * @return $this
+     * @since 1.9.0
+     *
+     */
+    public function register()
+    {
+
+        $assetFile = file_get_contents($this->getAssetFilePath());
```
New0, line 134: Should we check if file_exists() for $this->getAssetFilePath() ?

classes/api/forms.php
```diff
@@ -19,6 +19,31 @@ class Caldera_Forms_API_Forms extends  Caldera_Forms_API_CRUD {
...
 		parent::add_routes($namespace);
+        register_rest_route( $namespace, $this->id_endpoint_url(),
+           [
+                'methods'             => \WP_REST_Server::EDITABLE,
+                'callback'            => [ $this, 'save_form' ],
+                'permission_callback' => [ $this, 'save_form_permissions_check' ],
+                'args'                => [
+                    'cf_edit_nonce' => [
+                        'type' => 'string',
+                        'description' => __('Caldera Forms editor nonce'),
```
New0, line 30: We should add the 'caldera-forms' text-domain to __() function

classes/api/forms.php
```diff
@@ -19,6 +19,31 @@ class Caldera_Forms_API_Forms extends  Caldera_Forms_API_CRUD {
...
+                        'required' => 'true'
+                    ],
+                    'config' => [
+                        'type' => 'object',
+                        'description' => __('Caldera Forms editor nonce'),
```
New0, line 35: We should add the 'caldera-forms' text-domain to __() function

classes/api/forms.php
```diff
@@ -19,6 +19,31 @@ class Caldera_Forms_API_Forms extends  Caldera_Forms_API_CRUD {
...
+                        'required' => 'true'
+                    ],
+                    'form' => [
+                        'type' => 'string',
+                        'description' => __('ID of form'),
```
New0, line 40: We should add the 'caldera-forms' text-domain to __() function

classes/api/forms.php
```diff
@@ -57,7 +82,36 @@ public function add_routes( $namespace ) {
...
 
+    /**
+     * Save a form via REST API
+     * 
+     * Handler for POST /cf-api/v2/forms
+     * 
+     * @since 1.9.0
+     */
+	public function save_form(\WP_REST_Request $request){
+        $saved = Caldera_Forms_Admin::save_a_form(array_merge(['ID' => $request['form_id']],$request['config']));
```
New0, line 109: Should we check if $request['form_id'] and $request['config'] are set?

Shelob9:
>Before approval, I wanted to make sure that the conditions are not expected to work on frontend on this PR? And that will be treated seperately.

This is an incorrect assumption. Are they not working in the front-end?

Shelob9:
Commit b7e6f30dabfc8a5caa827bc5ecc6cfef7c4eda44 updated form-builder package to latest, which included adding the "OR" between conditionals and improved adding conditional - put focus on the group name input and I did not hide the button. 

Also, I am now updating form.fields[fieldId].conditions on save, so the conditonals work in the front-end.

classes/api/forms.php
Shelob9: Yes, we should. Fixed in b476c474

classes/api/forms.php
Shelob9: Fixed in b476c474

classes/api/forms.php
Shelob9: I don't think we need to beacuse:

- The request would not be valid without them, WordPress should refuse the request if they are not set.
- If they are not set correctly, ` Caldera_Forms_Admin::save_a_form` should return a WP_Error, which is expected on the next line.

cf2/Asset/Register.php
Shelob9: >From my logs $corePath seems to be set with an ending slash, clients should be set without first slash as on above line ( 'clients/' )

Correct

 I fixed that here: https://github.com/CalderaWP/Caldera-Forms/pull/3519/commits/d32dbb84a7b2fbe965c3d6feae1e414b2d9ec0ff#diff-f15134c609c4cc7436fa7e92019ad12fR67

cf2/Asset/Register.php
Shelob9: Yes.  It should always exist -- if assetts were built. Without it we can't register dependencies. So I will skip regesitration if it doesn't exist, which should be fixed by building dependencies.

Shelob9: See: 06fa04278ced2aff3ce77b0f2fc81d77c82f01fa

New0:
@Shelob9 I tested the latest commit and I found these points:
-  I was unable to make the conditions fully work, the fields are applied the condition when the form first loads but the condition is not applied during a change events on the state of the form / fields.

- The condition set in the Processor was not applied

- I was able to set a condition via the "Applied Fields" checkboxes options in the Condition view but couldn't from the Field's settings view via the "Condition" dropdown field. And no condition is applied to a field, the Condition dropdown shows "Disabled" as the option set.

Here is the field applied the condition in conditions view \:

Screenshot_2020-04-09 Caldera Forms Admin ‹ VSLAN — WordPress(1)
 
Here is the field settings view with a condition applied \:

Screenshot_2020-04-09 Caldera Forms Admin ‹ VSLAN — WordPress(3)

The field settings view without condition applied \:

Screenshot_2020-04-09 Caldera Forms Admin ‹ VSLAN — WordPress(4)

Shelob9:
@nico Thank you for the review.  This is super useful.

I'm _assuming_ we're going to clean up the CSS on the inputs and the new magic tag selector next week. Once we know this works and the PDF project is wrapped up.

I'm breaking this down into three parts:
- [ ] Conditonals are nto saved correctly : https://git.saturdaydrive.io/_/caldera-forms/tools/form-builder/issues/1
- [ ] https://git.saturdaydrive.io/_/caldera-forms/tools/form-builder/issues/2The type of processor conditonal is not saved.
- [ ] Conditonal type selector in the form field settings. I'm going to have to mess with edit.js. I wish I could make a reasonable case for removing that feature.

Shelob9:
@New0 I belive that all of the issues you mentioned before are now resolved as of 165da6f Can you please review again when you can?

If everything works as expected, we need to discuss CSS.

New0:
Thanks @Shelob9 , I could use the conditions in the form as expected but not the processors' conditions.
Did you remove the "Condition" dropdown select field intentionally from the "Field Settings" view?

Shelob9:
> Also, When I add a condition in a processor and reload the page, the processor apears as disabled.

@New0 is this the only issue functionality wise you see with this MR? The processor conditionals work as expected?

Shelob9:
@New0  re: CSS

I think we could make better use of horizontal space here:

spacing

The magic tag selector for conditionals is all new, I copied the CSS from the old one, its opening to the right of the field instead of under, which I don't dislike, but not with 100% width.

New0:
> 
> 
> > Also, When I add a condition in a processor and reload the page, the processor apears as disabled.
> 
> @New0 is this the only issue functionality wise you see with this MR? The processor conditionals work as expected otherwise?

Yes, I could see all other points reported working. 

@Shelob9 do you want me to work on the styles in this branch?

Shelob9:
> Yes, I could see all other points reported working.

Thanks for clarifying. 

>@Shelob9 do you want me to work on the styles in this branch?

Beacuse I'm also working in this -- trying to figure out why processors are disabling now -- it's probably best if you make a branch off of this branch. Also, the PR for that branch into this branch will give us a good place to discuss those changes.

Shelob9:
Issue for not saving the active state of processor: https://git.saturdaydrive.io/_/caldera-forms/tools/form-builder/issues/4

New0:
@Shelob9 after pulling the last commit, adding a new field to the Form breaks the interface. The "Save" button disappears and the "Add Field" button stays stuck where I dragged it.
After pulling I ran Yarn build.
I don't get any error in the debug.log and get the React minified error in the browser's console.
( Error: "Minified React error #200; visit https://reactjs.org/docs/error-decoder.html?invariant=200 for the full message or use the non-minified dev environment for full errors and additional helpful warnings. " )

Here is how the view stays stuck:
Screenshot_2020-04-29 Caldera Forms Admin ‹ VSLAN — WordPress

Shelob9:
@New0 Sorry I should have noted that a5e4276 will break everything I was 90% done with the next commit when I had to quit for the day. Can you please branch from the previous commit to work on style?

BTW We should discuss the boundries of form builder app in detail -- maybe a call tomorrow -- but the save button, and field selector in field settings are part of the React app now. See clients/form-builder/index.js

Shelob9:
@New0 After 7712656 adding new fields should not cause the error. Beacuse the selector in the field settings for conditional groups is a React component that is rendered using a portal on a node that is created by PHP and handlebars, I had to prevent rendering before that node exists. Took a minute, but I think I got it.

We should go over all of this on a screen share next week. I want to move the seperate form-builder repo into core plugin _after_ this is merged.

Shelob9:
@New0 After 7712656 adding new fields should not cause the error. Beacuse the selector in the field settings for conditional groups is a React component that is rendered using a portal on a node that is created by PHP and handlebars, I had to prevent rendering before that node exists. Took a minute, but I think I got it.

We should go over all of this on a screen share next week. I want to move the seperate form-builder repo into core plugin _after_ this is merged.

New0:
@Shelob9, after I ran yarn build, the elements are not showing on the edition page at all.  The save button and conditions under the conditions tab and processors are not loaded. I have no errors in console or logs.

Shelob9:
@New0 I just did `rm -rf node_modules && rm yarn.lock` then `yarn && yarn build` it still works for me. Can you try again please?

New0:
@Shelob9 the processor conditions settings are still not loaded. The condition from the conditions tab  all work as expected. Only the processor component is not loaded.

Shelob9:
> the processor conditions settings are still not loaded. The condition from the conditions tab all work as expected. Only the processor component is not loaded.

@New0 These issues should be fixed as of 25f50b00a6d4578385d6dabd25f7f19c09ece7f8

Shelob9:
- [ ] Validation is not running on save. This is beacuse the check was not copied from legacy to client https://github.com/CalderaWP/Caldera-Forms/blob/master/assets/js/edit.js#L107-L109

Shelob9:
@New0 Once #3540 is merged into this branch. I think we should make a build from this branch, call it CF 1.9.0 beta 1 and distribute that for public beta testing. We can start merging other PRs to develop. If beta testing goes well, we merge to develop and then make a beta 2 (RC1?) build with these fixes and those bug fixes.

Shelob9:
@New0 Thanks for mergin those changes.

Now I'm going back to your comment here: https://github.com/CalderaWP/Caldera-Forms/pull/3540#issuecomment-634682332

This suggests that wp-api-fetch is not enqueued. It should be beacuse of these lines of code should get the dependencies from the assets file, which I do see generated....

https://github.com/CalderaWP/Caldera-Forms/blob/master/classes/render/assets.php#L713-L716

Shelob9:
@Shelob9 `@wordpress/api-fetch` and `wp.apiRequest` are not the same thing

New0:
Thanks for fixing this @Shelob9 !

I ran more tests, the In-Form conditions are working as expected but not the Processors conditions.

I have three comments about the in-Processor conditions:

1. There isn't a blank state for the Type of Condition, currently, I only have two options, "Use" and "Don't Use". ( Note that currently, if not rules are set, the processor is triggered no matter which type is set )

2. When I set a condition  rule:

- With "Use" type of condition and a rule the processor is never used. ( The condition being met or not. )
- With "Don't Use' and a rule set, the processor is always Used. ( The condition being met or not. )

3. The 'Or' text between Rules is not displayed.

Shelob9:
@New0 Look it's a Caldera Forms 1.9.0 beta 1 ZIP. 

[1.9.0-b.1.zip](https://github.com/CalderaWP/Caldera-Forms/files/4704340/1.9.0-b.1.zip)

New0:
@Shelob9 I released a B2 with the current work merged to develop branch - 
[1.9.0-b2.zip](https://github.com/CalderaWP/Caldera-Forms/files/4765137/1.9.0-b2.zip) -

Shelob9:
> I released a B2 with the current work merged to develop branch -

@New0 Awesome.

Shelob9:
Caldera Forms 1.9.0 Release Candiate 1  [1.9.0-RC.1.zip](https://github.com/CalderaWP/Caldera-Forms/files/4766861/1.9.0-RC.1.zip)

PR #3527, opened 2020-04-23
Use DOM node construction vs concatenation for admin notice

jspellman814:


New0:
Hello @jspellman814 thank you for your suggestion. We're going to keep the concatenation to match what is on the current SendWP SDK.
Don't hesitate to keep helping.

jspellman814:
Sounds good @New0. In case you're interested in some additional context here, this was flagged as a security risk on the WordPress VIP platform. Specific error: `Error: HTML string concatenation detected, this is a security risk, use DOM node construction or a templating language instead: sendwp_vars+ (WordPressVIPMinimum.JS.StringConcat.Found).`

New0:
Thanks for the explanation @jspellman814 that makes sense.  I'm opening this as an issue on the SendWP sdk repo, and I'll probably open a new PR to cover your change.

Issue #3536, opened 2020-05-08
Error PHP 7.4


wesleyas92:
Deprecated: Unparenthesized `a ? b : c ? d : e` is deprecated. Use either `(a ? b : c) ? d : e` or `a ? b : (c ? d : e)` in C:\laragon\www\vinu-wp\wp-content\plugins\caldera-forms\classes\core.php on line 3732

New0:
Thanks for reporting this issue, this is a duplicate of #3476 
A PR was opened to address this and be reviewed for 1.9.0 milestone.

Shelob9:
This issue should be fixed in CF 1.9. a release candidate build, not fully tested can be downloaded here:

https://github.com/CalderaWP/Caldera-Forms/files/4766861/1.9.0-RC.1.zip

We welcome any testing, please let us know if this fixes your issue.

New0:
Closed via - https://github.com/CalderaWP/Caldera-Forms/commit/4cacd21dedbeb6db283d3785e500c34d9ba33d1f -

Issue #3476, opened 2020-01-30
Deprecated: Unparenthesized `a ? b : c ? d : e` (deprecated usage in PHP7.4)


budrick:
## What Is The Unexpected Behaviour?
Occasionally, I'll see the following error either logged (or appear on frontend and admin pages with WP_DEBUG and WP_DEBUG_DISPLAY enabled):

```
Deprecated: Unparenthesized `a ? b : c ? d : e` is deprecated. Use either `(a ? b : c) ? d : e` or `a ? b : (c ? d : e)` in (redacted)/wp-content/plugins/caldera-forms/classes/core.php on line 3724
```

This is due to the use of nested ternary operators and reliance on operator precedence rules without explicit parentheses. This usage of nested ternary operators without parens is deprecated in PHP 7.4, hence the notice. 

I haven't seen any other deprecation notices, but I suppose it doesn't mean they're not possibly there.

## What PHP Errors Have You Logged While Reproducing This Bug?
```
Deprecated: Unparenthesized `a ? b : c ? d : e` is deprecated. Use either `(a ? b : c) ? d : e` or `a ? b : (c ? d : e)` in (redacted)/wp-content/plugins/caldera-forms/classes/core.php on line 3724

**Comment by ralphjesy12** [comment]
I just quickly enclosed it with parenthesis. If there's anyone who wants to make the PR
`../caldera-forms/classes/core.php:3724`
From:
```
$entry_id = !empty($wp_query->query_vars[ 'cf_entry' ]) ? $wp_query->query_vars[ 'cf_entry' ]
: isset($_GET, $_GET[ 'entry' ]) && absint($_GET[ 'entry' ]) ? $_GET[ 'entry' ] : null;
```

To:
```
$entry_id = !empty($wp_query->query_vars[ 'cf_entry' ]) ? $wp_query->query_vars[ 'cf_entry' ]
: (isset($_GET, $_GET[ 'entry' ]) && absint($_GET[ 'entry' ]) ? $_GET[ 'entry' ] : null);
```

New0:
Thanks for reporting this issue, a PR was opened to address this and be reviwed for 1.9.0 milestone.

artshostak:
Having the very same issue, when can we expect the parentheses to be added as this makes the Wordpress dashboard entirely unusable at the moment:


[...]

PR #3539, opened 2020-05-11
Fixes deprecated warnings in PHP 7.4

ravigehlot:
The following deprecated warning comes up with PHP 7.4: default: Deprecated: Unparenthesized `a ? b : c ? d : e` is deprecated. Use either `(a ? b : c) ? d : e` or `a ? b : (c ? d : e)` in /home/vagrant/Projects/wp-content/plugins/caldera-forms/classes/core.php on line 3732. This PR encloses the first ternary.


New0:
Hello @ravigehlot , thank you for opening the PR. I am marking this to be reviwed for 1.9.0 milestone.

We will need to check if this issue could happen in other places and complete the PR if this is the case.

This works for #3476 and #3536

Issue #3531, opened 2020-04-29
Styles for form builder component


New0:
We want to adapt current placement of components for #3519

New0:
Closed via https://github.com/CalderaWP/Caldera-Forms/commit/c3b2055a51b4bdeb87db842e2bea104d7ae6d954

PR #3540, opened 2020-05-12
Form builder conditions styles

New0:
For #3531 

This adds styles to the admin stylesheet already enqueued. I used an @import in clients/admin/index.scss that imports styles/form-builder.scss

It improves spacing and positioning but we could move elements in the HTML to be more effective. In particular the magic-tags-list and the list of fields to assign the condition to.


New0:
@Shelob9 I tested before merging but couldn't save a form. I know this is not the purpose of this branch but as the source branch is saving correctly I don't want to introduce a bug in it and think we need to fix the issue here.

Here is where I am now \:

- pulled branch feature/3531-form-builder-styles
- removed /vendor, /node_modules, yarn.lock and composer.lock from my environment
- ran yarn && composer install
- rebuilt assets yarn build

Then when I visit the form builder, the styles are shown as expected but when I try to save the Form I get an error in the console, no error in PHP logs.

The console error is TypeError: s(...) is not a function => index.min.js l2526 which corresponds to l138 in form-builder/index.js , an apiFetch() call.

So far I tried (but it didn't help) \:

- removing the scss file import which is I think the only change in that file compared to source branch
- re-installing @wordpress/api-fetch dependency as I noticed it is set to "latest" version in package.json. (Reinstalling sets the version to 3.8.1)

My next step would be to look for an issue in the changes made to enqueue the assets as this is the last change made in this branch but I wanted to check with you before spending more time on this in case you know where it comes from.

Shelob9:
@New0 I think that the issues with API fetch are likely environment problems on your end. I didn't not have that issue. I did have a bunch of issues moving between branches and getting dependencies installed wrong.

Also, all of that is unrelated to styles. I think if you're happy with that the styles, we should merge this into the main MR and then move forward with testing functionality there.

Issue #2880, opened 2018-12-24
Feature - email address blacklist


Shelob9:
https://halfelf.org/2018/spam-your-blacklist/amp/?__twitter_impression=true

PR #3544, opened 2020-05-28
Use commnent black list for anti-spam 

Shelob9:
#2880

## What Has Changed

I created a new method in the Caldera_Forms class that gets the list of bad words from WordPress. It has a filter so end users can customize. I then copied logic from here https://gist.github.com/them-es/7c8a120ac868f5c4fc474e53a1266c83 to the loop where fields are validated.

## How to test

I'm not sure yet. But I think adding words to the block list and submitting form with those words in the fields.

* https://github.com/splorp/wordpress-comment-blacklist
* https://themeisle.com/blog/stop-comment-spam-on-wordpress/

### DIFF
--- a/classes/core.php
+++ b/classes/core.php
@@ -3096,12 +3096,29 @@ class Caldera_Forms
 			$future_fields = $_POST['_cf_future'];
 		}
 
+		//https://developer.wordpress.org/reference/functions/wp_blacklist_check/
+        //https://halfelf.org/2018/spam-your-blacklist/amp/?__twitter_impression=true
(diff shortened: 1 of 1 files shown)


Zodiac1978:
This needs a refresh after WordPress 5.5
See: https://make.wordpress.org/core/2020/07/23/codebase-language-improvements-in-5-5/

It still works, but on WP 5.5 it would trigger a deprecated warning.

There should be something like this:
```
if ( false === get_option( 'disallowed_keys' ) ) {
    // Assume this is WP < 5.5. Option does not exist.
} else {
    // Assume this is WP >= 5.5
}
```

Issue #3547, opened 2020-06-04
Update SendWP SDK


New0:
The SendWP connection needs to be adapted to last changes from the SDK.

New0:
Closed via https://github.com/CalderaWP/Caldera-Forms/commit/12282c9a7ad39d5336d60169d347acf4792f38a0

PR #3548, opened 2020-06-04
Update Sendwp sdk

New0:
For #3547 

This will integrate changes to fix the issue reported at #3527


Shelob9:
@New0 I just tested this locally. 

I clicked the blue SendWP button and was taken in to http://localhost:8228/wp-admin/undefined

I went back to cf admin page. From the console, I ran `window.caldera_forms_sendwp_remote_install()` and I was then taken to  http://localhost:8228/wp-admin/undefined

New0:
@Shelob9 do you have a local SendWP server environment declared in your CF local environment?

If not, clicking the "Signup for SendWP" button should open "https://sendwp.com/_/signup/register/"

But I know that I have set a local sendwp server using `define('SENDWP_SERVER_URL', 'https://sendwpserver.lndo.site');` and 

Also the local not being verified over https, it may require \:
`add_filter( 'https_local_ssl_verify', '__return_false' );`
`add_filter( 'https_ssl_verify', '__return_false' );`

Also in order to test this, it will require to click again on the button after being already connected to display an error.

Shelob9:
I do not have that constant set.

New0:
@josh did you use hooks to passby https vrification?
`add_filter( 'https_local_ssl_verify', '__return_false' );`
`add_filter( 'https_ssl_verify', '__return_false' );`

I'm not sure how to help you test this, the caldera_forms_sendwp_remote_install() js function triggers an ajax request that sets data in via the handler.php file in caldera_forms_sendwp_remote_install() php function.

The register_url that is used in the form.submit() in installer.js l57 gets defined in handler.php l88 which calls sendwp_get_server_url() function from the SendWP plugin.
The sendwp_get_server_url() function is located in the SendWP plugin in includes/functions.php l3

Shelob9:
@New0 

You asked -

```
 did you use hooks to passby https vrification?
add_filter( 'https_local_ssl_verify', '__return_false' );
add_filter( 'https_ssl_verify', '__return_false' );
```

No, I did not. I am unsure WHY I would need to add this beacuse I have NOT set the `SENDWP_SERVER_URL` constant, beacuse I do not know WHY I would need to, wouldn't we want to verify the SSL cert of SendWP.com before downloading?

I will test now with those filters.

Shelob9:
@New0 I tested again with `add_filter( 'https_local_ssl_verify', '__return_false' );
add_filter( 'https_ssl_verify', '__return_false' );` set in my mu-plugin. I am still taken to `/undefined` when I click the blue SendWP button.

sendwp/installer.js
```diff
@@ -3,24 +3,37 @@ function caldera_forms_sendwp_remote_install() {
...
         'sendwp_nonce': sendwp_vars.nonce
     };
+    
     // since 2.8 ajaxurl is always defined in the admin header and points to admin-ajax.php
     jQuery.post(ajaxurl, data, function(response) {
         var data = JSON.parse(response);
-        //Check for errors before calling caldera_forms_sendwp_register_client()
-        if(data.error === true ){
+         
+         if(data.error === true ){
```
Shelob9, line 11: @New0 In my local testing. The response is `{error: 'Something went wrong ..."}` when it fails, which explains why I am getting the redirect to `/undefined`. I think this could just be `if(data.errror){...`

Shelob9:
@New0 I just tested on a real site with SSL and such. It worked as expected.

I think that my testing shows that when this doesn't work, it doesn't fail well. See: https://github.com/CalderaWP/Caldera-Forms/pull/3548/files#r437457419 

That's a quick fix then we can deploy this.

New0:
@Shelob9 I'm glad you could test this! And the data error check makes sense without the boolean I tested both a successful connection and an errored request without the boolean and it works. - https://github.com/CalderaWP/Caldera-Forms/pull/3548/commits/77f5680f0a0d3a560fb10f049cf22028e314e876 -

Shelob9:
@New0 Happy to test. It took me a minute to figure out it was never going to work with my local CF + actual SendWP.com, but happy it exposed the edge case. Your last commit 77f5680f0a0d3a560fb10f049cf22028e314e876 looks like it will fix that. Thanks for working with me on this.

PR #3566, opened 2020-06-22
 Impliment basic form cache 

### LINKED ISSUE #3553: Duplicate queries selecting from form table on every request.
Short Debug Information:

`WordPress Version: 5.4.2 PHP Version: 7.3.18 MySQL Version: 5.5.5 Caldera Forms Version:   1.8.11`

Currently Active Theme: Twenty Twenty: 1.2
Akismet Anti-Spam: 4.1.5 Not Active
Caldera Forms: 1.8.11 Active
Hello Dolly: 1.7.2 Not Active
Query Monitor: 3.6.0 Active
Yes.

I read these reports: 
https://wordpress.org/support/topic/duplicate-queries-10/ and I looked at CalderaForms.com new relic, which shows wp_cf_forms select as slowest query for last 7 days and last 60 minutes.

On  site with no forms, on front-end page with no forms, I see a duplicate query in Query monitor from `wp_cf_forms`

dupes

Adding a form changes the duplicate query

dupes-2

Thriplicated queries on the page with form:

image

None

none

Shelob9:
Fixes #3553 

## What Has Changed

* When querying for form, I put form in a variable, use that instead of query if needed.
* Tests showing that duplicate queries are not run.
* Specifics detailed here: https://github.com/CalderaWP/Caldera-Forms/pull/3566#issuecomment-648293099

## How To Test

* Make sure tests pass
* Check in wp-admin, page without forms and page with form for duplicate queries using Query Monitor plugin

### DIFF
--- a/classes/db/form.php
+++ b/classes/db/form.php
@@ -248,9 +248,6 @@ class Caldera_Forms_DB_Form extends Caldera_Forms_DB_Base {
 	    }
 
 	    return $forms;
-
-
-
     }
 
 	/**
--- a/classes/forms.php
+++ b/classes/forms.php
@@ -235,13 +235,13 @@ class Caldera_Forms_Forms
      */
     protected static function get_stored_forms()
     {
-
         if (empty(self::$stored_forms)) {
             $forms = Caldera_Forms_DB_Form::get_instance()->get_all();
             if (!empty($forms)) {
                 foreach ($forms as $form) {
                     if (!empty($form['config'])) {
                         static::$stored_forms[$form['form_id']] = $form['form_id'];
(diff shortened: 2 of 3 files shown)


Shelob9:
@New0 This is now ready for review.

- b4cc053 Added a cache on the query for each form, and prevents the duplicate query notice in query monitor. But, it was still doing one query for all forms and then one query per form, and running 1 or more query on each page.  Better, but still (1+number of forms) queries per page.
- 0a013b5 Uses the query for all forms to populate the cache used when querying for each forms. With this, in query monitor I see 1 or 0 queries for the forms table, except in CF admin, where there are more, but that's expected.

Issue #3574, opened 2020-07-03
Test Github actions


New0:
Test converting the Gitlab CI into Github actions

New0:
Closed via https://github.com/CalderaWP/Caldera-Forms/pull/3575

PR #3575, opened 2020-07-03
Run tests in Github actions

New0:
For #3574 

Ran tests with same commands as we do locally instead of applying old actions


Shelob9:
@New0 I made these changes:

- Allowed e2e workflow to fail. It would be awesome if we can fix this. I might have time, but also priorities. https://github.com/CalderaWP/Caldera-Forms/issues/3583
- I deleted the Gitlab CI file. It's redundant.

I'm really glad you figured this stuff out. Awesome sauces, side of crispy tofu.

Issue #3571, opened 2020-06-26
No entries displayed with caldera_forms_entry_viewer-shortcode and 404 on call to REST API


perikorese:
I try to use the frontend shortcode for viewing entries, but I get "No Entries To Display". There are entries on the form.

## What Version Of Caldera Forms, WordPress and PHP Are You Using?
WordPress Version: 5.4.2
PHP Version: 7.3.5
MySQL Version: 5.7.28
Caldera Forms Version: 1.9.1
WP_DEBUG:

## Does Your Issue Persist When You Disable All Other Plugins and Switch To The Default Theme?
Yes.

## What Is The Unexpected Behaviour?
On a page in the frontend, I insert this shortcode: [caldera_forms_entry_viewer id="CF5ef5eb6a8ae49"]
I expect to see the entries that the form has.
But instead of entries I get this message: "No Entries To Display"

## What PHP Errors Have You Logged While Reproducing This Bug?
No errors.

## What JavaScript Errors Have You Seen While Reproducing This Bug?
I see that there is a call to the REST API, which results in a 404. Screenshot attached. I get this on a local dev site and also on a non-local staging site.
![Skærmbillede fra 2020-06-26 14-48-13](https://user-images.githubusercontent.com/52708770/85858577-2aeaff80-b7bc-11ea-813e-367cbbcad809.png)

I am logged in as administrator when viewing the page.
Caldera Forms is the only plugin installed.
The theme is "Twenty Twenty".

Shelob9:
I don't know why this is happening, but the request is being made to `/wp-json/cf-api/v2/forms/<form-id>` which is not a route that exists. It should be making a request to `/wp-json/cf-api/v2/entries/<form-id>`

riverBanjo:
https://desk.zoho.com/support/saturdaydriveinc/ShowHomePage.do#Cases/dv/413570000004636140

Shelob9:
@Shelob9 RE: https://github.com/CalderaWP/Caldera-Forms/issues/3571#issuecomment-652026360

ACTUALLY, the requests are correct. `/wp-json/cf-api/v2/forms/<form-id>` should be requested, it is not registed, **that is the actual bug.** 

CF 1.9.0 registered a POST endpoint on the same route as that GET request, which is probably the problem here.

Shelob9:
Should be resolved via #3576

New0:
[...]

Issue #3570, opened 2020-06-25
Form Saving Broken After 1.9.0 Update


GreenRooster:
## Do You Need Immediate Help?

N/A. Rolled back to working 1.8.11

## Is This A Feature Request?

No

## What Version Of Caldera Forms, WordPress and PHP Are You Using?

**Short Debug Information**
    WordPress Version: 5.4.2  |  PHP Version: 7.3.18-1+ubuntu18.04.1+deb.sury.org+1  |  MySQL Version: 5.7.29  |  Caldera Forms Version: 1.9.1  |  WP_DEBUG: 1

## Does Your Issue Persist When You Disable All Other Plugins and Switch To The Default Theme?
Yes

## What Is The Unexpected Behaviour?

Upon updating from 1.8.11 to 1.9.0 & 1.9.1, we have been unable to save either newly created forms or existing forms. We use post names for our permalinks. If we change permalinks to plain, things function properly. However, that is not an acceptable solution. Seems like a similar error to #3568.

In a fresh install on another testing site with the below debug settings, everything works as normal. Which is great, except I can't change some of what my host allows.

WordPress Version: 5.4.2  |  PHP Version: 7.4.1  |  MySQL Version: 8.0.16  |  Caldera Forms Version: 1.9.1

## What PHP Errors Have You Logged While Reproducing This Bug?

None.

## What JavaScript Errors Have You Seen While Reproducing This Bug?

Request URL: ***/wp-json/cf-api/v2/forms/undefined?_locale=user
Request Method: POST
Status Code: 404 Not Found

code: "invalid_json"
message: "The response is not a valid JSON response."

New0:
Than you for reporting this issue @GreenRooster 

@Shelob9 I could reproduce an issue saving the form when testing PR #3576 ( 

The investigation so far resulted in:

1. The API route set in classes/api/forms.php is not overriding the route from classes/api/crud.php
I believe the route set in classes/api/forms.php should be targetted in order to call save_form($request) l116

It is not the case in PR #3576 where clicking the Save button in a Form edition page targets the Route in classes/api/crud.php and calls update_item( $request) l254 which is currently set to not_yet_response(); ( as a route from crud ).

2.  The handleSave() in clients/form_builder/index.js l71 is called without a form_id which results in calling an API route as "forms/undefined"

I opened  PR #3581 to illustrate this issue combined with the fix for Privacy settings and entries edition.

andyburnsco:
[...]

Issue #3568, opened 2020-06-24
Using Plain Permalinks Breaks the Entry Editor


riverBanjo:
## Replication steps

1. Change Permalink settings to plain ( ?p=123 )
2. Install Caldera Forms
3. Make a form
4. Submit the form
5. View entry
6. Edit entry
7. See that the entry editor is covered by a home page

## Debug Info
WordPress Version: 5.4.2

PHP Version: 7.3.19

MySQL Version: 5.7.30

Server Software: Apache/2.4.38 (Debian)

Your User Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/83.0.4103.116 Safari/537.36

Session Save Path:

Session Save Path Exists: No

Session Save Path Writeable: No

Session Max Lifetime: 1440

Opcode Cache:

Apc: No
Memcached: No
Redis: No
Object Cache:

Apc: No
Apcu: No
Memcache: No
Memcached: No
Redis: No
WPDB Prefix: wp_

WP Multisite Mode: No

WP Memory Limit: 40M

Currently Active Theme: Twenty Nineteen: 1.5

Plugins:

Akismet Anti-Spam: 4.1.6 Not Active
Authorize.net for Caldera Forms: 2.1.4 Active
Caldera Forms: 1.9.1 Active
CalderaWP License Manager: 1.2.10 Active
MailPoet 2: 2.13 Not Active
Ninja Forms: 3.4.24.3 Not Active
Ninja Forms - ActiveCampaign: 3.0.6 Not Active
Ninja Forms - Addon Manager: 3.0.13 Not Active
Ninja Forms - AWeber: 3.1.1 Not Active
Ninja Forms - Campaign Monitor: 3.0.5 Not Active
Ninja Forms - Capsule CRM: 3.4.0 Not Active
Ninja Forms - CleverReach: 3.1.3 Not Active
Ninja Forms - ClickSend: 3.0.1 Not Active
Ninja Forms - Conditional Logic: 3.0.26 Not Active
Ninja Forms - Constant Contact: 3.0.4 Not Active
Ninja Forms - ConvertKit: 3.0.2 Not Active
Ninja Forms - Elavon Payment Gateway: 3.1.0 Not Active
Ninja Forms - EmailOctopus: 3.0.0 Not Active
Ninja Forms - Emma: 3.0.4 Not Active
[...]

PR #3576, opened 2020-07-03
Register missing rest api routes 

Shelob9:
#3571

* In cf 1.8.x  there were 5 HTTP methods for `/cf-api/v2/forms/(?P<form_id>[\w-]+): ` route
* In CF 1.9.0+ there are 3 methods
* You can not GET a form via API, so entry viewer doesn't work

## What Has Changed

* All 5 endpoints are registered
    - Go to route discovery for cf-api/v2 at /wp-json/cf-api/v2 and check `/cf-api/v2/forms/(?P<form_id>[\w-]+): ` route for number of methods
* Entry viewer shortcodes like `[caldera_forms_entry_viewer id="CF59d957b7acff7"]` work as expected.


New0:
@Shelob9 I could test and confirm this fixes Privacy settings issue #3577 but I had an issue saving the Form and think it's related to the same API routes settings, I described the issue in #3570

New0:
@Shelob9 this PR now fixes privacy settings #3577and entry viewer shortcode #3571 and I can't reproduce #3570 anymore ( I use this formulation because I'm not sure the original issue reported was caused by the same code, but I can't reproduce it anymore from this branch ).

Shelob9:
@New0 Awesome. This is close.

Here is the litteral problem with plain permalinks -- I assumed pretty permalinks here https://github.com/CalderaWP/Caldera-Forms/blob/master/clients/form-builder/index.js#L153

New0:
@Shelob9 after running tests, the precise API call you linked worked as expected using Plain permalinks, the URL is converted to 

POST | https://local.lndo.site/index.php?rest_route=/cf-api/v2/forms/CF5eb5616ea264f&_locale=user

it means we can save forms as expected with all permalinks settings.

But setting plain permalinks breaks Privacy settings and Entries shortcode as there are query variables that create an URL with two "?"

For example, selecting a Form in the Privacy settings page create this URL 

GET | https://local.lndo.site/index.php?rest_route=/cf-api/v2/forms/CF5ef20e90ddd7f?privacy=true

we can see the second "?" which I think causes the issue.

Or the entries shortcode API call that results in:
GET | https://local.lndo.site/index.php?rest_route=/cf-api/v2/entries/CF5eb5616ea264f?page=1&per_page=80

Shelob9:
@New0 Nice finds. I super appreciate the extra attention to detail, I over simplified.

For the entry viewer, the extra page/per page args are added to url here: 
https://github.com/CalderaWP/Caldera-Forms/blob/efa47e26c09c8b528c17493ee1126506d1162228/assets/js/entry-viewer-2.js#L58-L64

For privacy settings, look here: https://github.com/CalderaWP/Caldera-Forms/blob/2cdb95b7f258b6567f0b570da526ab4bd6d58ec3/clients/state/api/index.js

For entry viewer AND privacy client, the rest api route is passed to DOM using the same config. See https://github.com/CalderaWP/Caldera-Forms/blob/2cdb95b7f258b6567f0b570da526ab4bd6d58ec3/classes/api/jsconfig.php#L68-L80

Shelob9:
@New0 3074b0c9d037b1535bee92b38a5ec0946fb83621 addresses this for entry viewer. The bug for privacy settings editor is in another library. It was my intent to merge all that into core using Yarn workspaces, so this sort of situation wouldn't be a pain. But priorities.

- https://github.com/CalderaWP/Caldera-Forms/blob/3571-fix-get-forms-api/clients/state/api/apiClients.js#L1 
- https://github.com/CalderaWP/caldera-api-client/blob/master/src/EntriesClient.js#L16-L21

New0:
@Shelob9 I can confirm I was able to Save forms, Edit entries in the entry viewer, Set privacy settings, use the Entry viewer shortcode as expected on this branch.

The only fix left will happen in an external repository to fix Privacy settings using Plain permalinks.

Shelob9:
> @Shelob9 I can confirm I was able to Save forms, Edit entries in the entry viewer, Set privacy settings, use the Entry viewer shortcode as expected on this branch.

OK great. Thanks for confirming that. That is what I expected.

> The only fix left will happen in an external repository to fix Privacy settings using Plain permalinks.

That's likely not happening. I don't want to mess with that stuff before the cleanups I proposed here: https://3.basecamp.com/4261010/buckets/13143498/messages/2761620839 but we're not going to do any of that right now.

Maybe after we import the module we can replace that function on the main object with a  similar fix for URL generation as I used for entry viewer? 

Also, the feature works, just not with an uncommon permalink settings. Maybe it is best to release CF 1.9.2 with all these big fixes and open a new issue for this edge case?

Shelob9:
@Shelob9 Fine, but also, it's one line that needs changed here: https://github.com/CalderaWP/caldera-api-client/blob/master/src/ApiClient.js#L254

Shelob9:
> Maybe after we import the module we can replace that function on the main object with a similar fix for URL generation as I used for entry viewer?

Should be a good quick fix and work. After c26c419 I see the new console log, I can fix this.

Shelob9:
@New0 This should be fixed now.

- a508684 - Use a closure function, not an arrow function so that `this` is defined in said function.
- 6b9883a - Use that function for both clients in a way that should solve  #3577

New0:
@Shelob9 I tested routes for Saving form settings - Saving Privacy settings - Editing the Entry via entry viewer - Display entries with Entry viewer shortcode.

**Using Pretty Permalinks** everything worked but the Entry viewer shortcode. The call below resulted in  an Error status 404 "No route was found matching the URL and request method"

GET | http://local.lndo.site/wp-json/cf-api/v2/entries/CF5eb5616ea264f&page=1&per_page=21

{"GET":{"scheme":"http","host":"local.lndo.site","filename":"/wp-json/cf-api/v2/entries/CF5eb5616ea264f&page=1&per_page=21","remote":{"Adresse":"127.0.0.1:80"}}}

**Using Plain Permalinks** I could still reproduce #3568 , I shared details in the issue.

I think we need to fix the Entry Viewer issue using Pretty permalinks for 1.9.2, fixing the plain permalinks issue for editing entry would be a plus.

Shelob9:
Possibly a better global solution for creating URLs:


Right now, we're doing a lot of checking strings for ? and &, instead of using the URL object. Something like this:

```js
function(endpoint,args={}){
  //parse url
  //https://developer.mozilla.org/en-US/docs/Web/API/URL
  let url = new URL(endpoint);
  //Add params
  Object.keys(args).forEach( arg => {
      //https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/URLSearchParams
      url.searchParams.append( arg, args[arg] );
  });
  
  return url;

}
```

I presume these tests would pass for that:
```
it( 'deals with a ? in the url', () => {
let url = addArgs( 'https://nico.com?rest_route=forms', {id:'cf3'});
expect( url ).toBe( 'https://nico.com?rest_route=forms&id=cf3' );
});

it( 'deals with no ? in the url', () => {
let url = addArgs( 'https://nico.com/forms', {id:'cf3'});
expect( url ).toBe( 'https://nico.com/forms?id=cf3' );
});
```

Then we could drop that in -- using cut in paste for assets/js -- to all url generating functions.
```

**Comment by amelsens** [comment]
Hello,
I'm sorry to interrupt you....
I could not help you with PHP code. I'm not an expert programmer. 
But if I can help with online tests on Linux webservers (via Hosting) or on my different Windows local IIS webservers or even on a Amazon virtual server online, don't hesitate to ask me that.
Maybe if you will tests some 1.9.2 beta versions, no problems for me.
Please, let me know something about it. I will be glad if I can help. ;)

**Comment by New0** [comment]
@Shelob9 I changed how we use indexOf to set the correct separator as it doesn't return a boolean but the position number if found and -1 if not found. 

@amelsens Thank you for your help, here is a beta - [1.9.2-beta1.zip](https://github.com/CalderaWP/Caldera-Forms/files/4918837/1.9.2-beta1.zip) - 
Could you run manual tests from your environment and make sure the Privacy settings and the Form settings are saved as expected?

**Comment by Shelob9** [comment]
> @Shelob9 I changed how we use indexOf to set the correct separator as it doesn't return a boolean but the position number if found and -1 if not found.

@New0 Good find. I just ran this in the console to remind my self why you are correct:

```js
let isNegativeOneTrue = -1 ? true : false;
//It is.
```

I think in my mind I assume that -1 is falsey, but it's truthy. Looks strange, but 1 is truthy and -1 is 1 with its polarity reversed.

New0:
This is a beta2 that includes the fix for editing entries using Plain permalinks structure - 
[1.9.2-beta2.zip](https://github.com/CalderaWP/Caldera-Forms/files/4919468/1.9.2-beta2.zip)

amelsens:
Hello, 
Sorry I missed up already Beta1 but I tested Beta2 and the Privacy Settings are working now. No error message anymore!
After changing other things I have also changed some Privacy Settings and they are saved as expected.
I will test again and again... And if I see something that's not normal, I will let you know.

amelsens:
Hello,
I have a strange behaviour when I want to disable one field from the GDPR fields settings...
When I click on the enabled Yes button for example, for the last field, first all the other fields became automatically disabled and not the field I indicate. I must click twice to disabling it. And then, I need to enabling all the other fields again.
Is this normal? I guess that when I need to disable only 1 field, that the other fields may stay unchanged. The may keep there status as saved, so in my case enabled.
See also the screencaptures
Privacy Settings ‹ clicking 1 field   previous fields auto disabled
Privacy Settings ‹ disabling 1 field   previous fields enabling again
Privacy Settings ‹ all fields enabled


 in attachment.

New0:
Hello @amelsens thank you very much for testing the beta. I opened a new issue based on your comment about the strange behaviour in Privacy settings, this will need to be treated separately.

amelsens:
Hello @New0 
I'm very glad I could help. 
Please let me know if there is a new Beta version.

Issue #3578, opened 2020-07-06
Update phpunit to 7.4


Shelob9:
This should be a part of PR #3574

Shelob9:
@New0 CF uses phpunit 6, everything else is using phpunit 7.3~5. It would be good if CF used phpunit 7.x or so. 

Also, given that you figured out how to make Caldera Forms' docker-compose file work for Github actions, it would be super cool if we could just copy and paste all that into Ninja Forms  -- but Ninja Forms needs phpunit 7.x. 

So, I'm going to try and update this in CF, then what you're doing in #3574 + this gives us everything we need for Ninja Forms.

Shelob9:
Close via #3579

PR #3579, opened 2020-07-06
Upgrade to phpunit 7 for testing in docker-compose and CI

Shelob9:
#3578

## What Has Changed

* phpunit for unit and integration tests is now ^7.3. 7.5 is being used, which is great.
     - https://github.com/CalderaWP/Caldera-Forms/pull/3579/files#diff-47bc3f2d114cb29ed9664b2bb82d0626
* We now have our own dockerfile for that container. That will be slow. It should get published to hub or the organization's docker registery. That will make things faster and also re-usable. For now, it's fine.

## How to test

- Includes #3575 for #3574 so, if pipeline passes, it's good.
- Also, test by checking out locally and running tests as documented.

In both cases, ensure php 7.3 or later is being used.


Shelob9:
I think this "fail" is good enough. https://github.com/CalderaWP/Caldera-Forms/pull/3579/checks?check_run_id=842690668

Two reasons it does not exit sucessfully:

- Math functions triggering deprecation notices that we expect to happen.
- One empty test.

image

Shelob9:
b33a8e922f5d59bc94393c18e72ac0f35ef37bf6 makes tests pass. Learned alot from : https://github.com/sebastianbergmann/phpunit/issues/3758

PR #3581, opened 2020-07-07
Save Forms via API

New0:
I could reproduce an issue saving the form when testing PR #3576 described in issue #3570. 

This is a branch created from #3576 and the PR is created against #3576. 

This just intends to illustrate the issue and makes saving form process work (along with privacy settings and entries edition) but I believe this is not the intended way of having the Forms saved.

### DIFF
--- a/classes/api/crud.php
+++ b/classes/api/crud.php
@@ -253,9 +253,7 @@ abstract class Caldera_Forms_API_CRUD implements Caldera_Forms_API_Route
      */
     public function update_item(WP_REST_Request $request)
     {
-
-        return $this->not_yet_response();
-           
+        return $this->not_yet_response(); 
     }
 
     /**
--- a/classes/api/crud.php
+++ b/classes/api/crud.php
@@ -254,7 +254,7 @@ abstract class Caldera_Forms_API_CRUD implements Caldera_Forms_API_Route
     public function update_item(WP_REST_Request $request)
     {
 
-        \Caldera_Forms_API_Forms::save_form( $request );
+        return $this->not_yet_response();
            
     }
 
--- a/classes/api/crud.php
+++ b/classes/api/crud.php
@@ -253,7 +253,7 @@ abstract class Caldera_Forms_API_CRUD implements Caldera_Forms_API_Route
      */
     public function update_item(WP_REST_Request $request)
     {
-        return $this->not_yet_response(); 
+        return $this->not_yet_response();
     }
 
     /**
--- a/classes/api/crud.php
+++ b/classes/api/crud.php
@@ -253,7 +253,9 @@ abstract class Caldera_Forms_API_CRUD implements Caldera_Forms_API_Route
      */
     public function update_item(WP_REST_Request $request)
     {
-        return $this->not_yet_response();
+
+        \Caldera_Forms_API_Forms::save_form( $request );
+           
     }
 
     /**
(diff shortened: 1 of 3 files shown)


classes/api/crud.php
```diff
@@ -253,7 +253,9 @@ public function create_item(WP_REST_Request $request)
      */
     public function update_item(WP_REST_Request $request)
     {
-        return $this->not_yet_response();
+
+        \Caldera_Forms_API_Forms::save_form( $request );
```
Shelob9, line 257: @New0 This is a base class extended by entries and forms, so it can't just save a form.

The reason I didn't extend this method in the forms API is that there was no way to change the request arguments.


- https://github.com/CalderaWP/Caldera-Forms/pull/3576/files#diff-1e2e9b0f9918e8d326a8e6306d158274R397
- https://github.com/CalderaWP/Caldera-Forms/pull/3576/files#diff-50f33cdd5ab83a6c8d8a844c07015635R32

clients/form-builder/index.js
```diff
@@ -69,6 +69,12 @@ const FieldConditionalSelectors = () => {
  * @since 1.9.0
  */
 const HandleSave = ({ jQuery, formId }) => {
+
+	//Reset formId if undefined
+	if(typeof formId === "undefined"){
+		formId = CF_ADMIN.form.ID;
```
Shelob9, line 75: @New0 This should be the only thing that needs fixed really.

New0:
> 
> 
> @New0 The real bug in #3570 is that the form ID in the save request is undefined. You addressed that here: https://github.com/CalderaWP/Caldera-Forms/pull/3581/files#r450895281 I would approve this PR if that was all it did.
> 
> Changes to the REST API are handled via #3576

Thank you for the review @Shelob9 , this branch was forked from PR #3576 branch, it includes the REST API fix. 
What I meant by the "unintended way" is that the route set in classes/api/forms.php that should override the one from classes/api/crud.php is not used when saving forms.

The Save Form process still calls update_item() in api/crud.php instead of save_form() in api/forms.php.

Shelob9:
> What I meant by the "unintended way" is that the route set in classes/api/forms.php that should override the one from classes/api/crud.php is not used when saving forms.

I understand that. I want you to remove the changes to that file and then merge this PR into branch `3571-fix-get-forms-api` for the following reasons:

- It is my intent for that method for the method to look like this:

```php
    public function update_item(WP_REST_Request $request)
    {
        return $this->not_yet_response();
    }
```

- If we need to change the REST API more before merging branch `3571-fix-get-forms-api`, we can, but right now, it has not I do not see why we would need to. https://github.com/CalderaWP/Caldera-Forms/pull/3581#discussion_r450894675

classes/api/crud.php
Shelob9: @New0 you wrote 

> The Save Form process still calls update_item() in api/crud.php instead of save_form() in api/forms.php." 

This statement is correct, and is what I intened. It smells bad, but it's a work around for a decision I made years ago that in hindsight wasn't great, and it works.

New0:
@Shelob9 I removed the unintended changes. This PR was a way to illustrate the issue but now that we are mentioning fixing #3570 here I add that this change alone doesn't fix #3570 , I'm marking this as a WIP.

Shelob9:
Sounds good. Once the extra change is out, please merge it right away. Don't ask for approval. That's just not needed it's a PR against a PR. That way we can close #3570  via  PR #3576

New0:
@Shelob9 I found the re-registration of "forms" routes in classes/api/forms.php l52 was setting routes with original values from $this->id_route(); instead of the edited route from $di_route.

I'm merging this in order to test Saving Forms + privacy settings via PR #3576

Issue #3590, opened 2020-07-15
Automatically generate a ZIP when a tag or release is generated


Shelob9:
When a tag is applied:

- Install and build for release
- ZIP files that we ship
- Attatch to release

Use this as reference:
https://github.com/Shelob9/actions-test/blob/master/.github/workflows/uploader.yml

Related #3575

New0:
Closed via - https://github.com/CalderaWP/Caldera-Forms/commit/40696f7e12328e06cd4c169762e5b019ac8a0ed8 -

PR #3591, opened 2020-07-15
Automate building zip files on when tagged

Shelob9:
#3590

## What has changed
Added a new workflow that when a tag is added or release is added, a ZIP file is generated and then added to the release page.

## How To Test
Not sure, I would have thaught I needed to do these two things:

- Add a tag
- Add a release on this branch.

Then test that the ZIP is uploaded correctly and has all of the files we need, but I can't trigger the action either way. Maybe it needs to be in develop first?

### DIFF
--- a/.github/workflows/zip.yml
+++ b/.github/workflows/zip.yml
@@ -1,0 +1,41 @@
+name: ZIP
+
+push:
+    tags:
+      - "*"
+  release:
+    types: [published]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Install PHP dependencies, optimized
+      run: composer install --no-dev -o
+
+    - name: Install JS dependencies
+      run: yarn
+
+    - name: Build assets for release
+      run: yarn build
+    # MUST set in package.json to same as tag name before pushing tag
+    - name: Set version number 
+      run: yarn version:number
+    - name: Create build
+      run: yarn package
+    - name: Create ZIP
+        uses: papeloto/action-zip@v1
+        with:
+          files: assets/ cf2/ classes/ fields/ languages/ vendor/ processors/ sendwp/ ui/ caldera-core.php readme.txt LICENSE
+          recursive: false
+          dest: caldera-forms-${{ github.ref }}.zip
+    - name: Upload ZIP to release
+      uses: svenstaro/upload-release-action@v2
+      with:
+        repo_token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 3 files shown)


Shelob9:
> Also note that the zip is giben the name of the tag - not the caldera-forms-${{ github.ref }}.zip name set in the configuration -
U did note that. I don't know why it's happening, but it is.

> And a cool download link is shown next to the tag in the tags list - https://github.com/CalderaWP/Caldera-Forms/tags -

I did and that is super awesome.

Shelob9:
@New0 Actually problem with this: built clients are NOT included. Either `clients/*/build/` if it works with globs or the functional equivent needs to get added to the ZIP actions list of files to include.

.github/workflows/zip.yml
```diff
@@ -0,0 +1,45 @@
...
+      - name: Set version number 
+        run: yarn version:number
+
+      - name: Create build
+        run: yarn package
+
+      - name: Create ZIP
+        uses: papeloto/action-zip@v1
+        with:
+          files: assets/ cf2/ classes/ fields/ languages/ vendor/ processors/ sendwp/ ui/ caldera-core.php readme.txt LICENSE
```
Shelob9, line 36: @New0 Hold on, if we run yarn package we have everything we need at build/version-number. Wouldn't it be simpler to ZIP that dir?

New0: @Shelob9  That would replicate the exact way we're building for production locally

Shelob9: Right, that seams better.

New0:
@Shelob9 I got it working with the build folder using the ${{ github.ref }} but I had to change the grunt task to paste the build at `build/refs/tags/<%= pkg.version %>` instead of `build/r<%= pkg.version %>`

I also run the action only on TAG pushed, running the same action on TAG push and Release published resulted in an error as a Release pushes a TAG and one would result in an error.

The process would require checking the version set in package.json and use the same string as the TAG or the TAG of the Release.

For example current version set is 1.9.2-zipTest :

- Go to https://github.com/CalderaWP/Caldera-Forms/releases

- Click the "Draft a new Release" button

- Set 1.9.2-zipTest as the TAG and select current branch 3590-zip

- Set a Title to the Release and a description

- Click the "Publish Release" button

- Wait for the action with the name of the Release to complete at  - https://github.com/CalderaWP/Caldera-Forms/actions - to complete

- Download ZIP from https://github.com/CalderaWP/Caldera-Forms/releases

In this case - https://github.com/CalderaWP/Caldera-Forms/releases/tag/1.9.2-zipTest -

Shelob9:
@New0 We should merge this, right?

Issue #3644, opened 2020-12-16
React warnings logged


New0:
These are the warnings to fix :

- Warning: Each child in a list should have a unique "key" prop.

- Warning: Unsupported style property vertical-align. Did you mean verticalAlign?

- Warning: Invalid DOM property `class`. Did you mean `className`?

New0:
This was fixed via - https://github.com/CalderaWP/Caldera-Forms/pull/3674 -

PR #3645, opened 2020-12-16
Fix react warnings

New0:
For #3644 

- [x] Warning: Each child in a list should have a unique "key" prop.

- [x] Warning: Unsupported style property vertical-align. Did you mean verticalAlign?

- [x] Warning: Invalid DOM property class. Did you mean className?

- [x] Warning: Received `true` for a non-boolean attribute `lazy`.


New0:
Included in PR #3674

Issue #3642, opened 2020-12-10
Form Creation does not work in WordPress 5.6


martin-neumann-gurus:
## What Version Of Caldera Forms, WordPress and PHP Are You Using?
*You can copy the "Short Debug" line from the Caldera Forms support menu's debug tab
WordPress Version: 5.6PHP Version: 7.3.25MySQL Version: 5.7.26Caldera Forms Version: 1.9.2WP_DEBUG:

## Does Your Issue Persist When You Disable All Other Plugins and Switch To The Default Theme?
Yes

## What Is The Unexpected Behaviour?
* Please explain what you did, what you expected to happen, and what happened instead.
I tried on several sites, when I update to WordPress 5.6, I cannot create new forms, even though it worked fine before the WordPress update.
I choose a name, but the create button does not work.

Renewrmk:
Install the Plugin "Enable jQuery Migrate Helper" and use older jquery Version will help.

KilakwaBT:
I am seeing this issue and we are not able to clone or export a form. I installed "Enable jQuery Migrate Helper". If I leave the settings as "Default from WordPress" and set to "Allow website visitors to trigger an automatic downgrading to legacy jQuery, when a failure is detected" the issue remains. If I set the jQuery version to "Legacy 1.12.4-wp" I can clone and create forms. My concern is this causes other issues.

Renewrmk:
"The Caldera Forms team has been working on the updates for the WordPress 5.6 upgrade to jQuery 3. We have the fixes but haven’t completed migrating and testing them in the codebase.
For the time being, please revert WordPress to a 5.5.x version and we will let you know as soon as we get our update out.
Thank you for your patience."

https://wordpress.org/support/topic/conflict-with-wordpress-version-5-6/

New0:
Partial notes while debugging this issue:

- Loading the Caldera Forms -> Forms page using jQuery 3.5.1 . The id ( normally set as 'id="baldrick_trigger_1608031375178"' ) of the "Create Form" button in the "Create New Form" Modal Form is not being set.
![Screenshot_2020-12-15 Caldera Forms Admin ‹ VSLAN — WordPress](https://user-images.githubusercontent.com/16348693/102209027-7d39ed00-3ed0-11eb-9e0d-cd9e14d15677.png)

[...]

Issue #3639, opened 2020-11-21
Export feature is broken on WordPress 5.6 RC1


Zodiac1978:
## What Version Of Caldera Forms, WordPress and PHP Are You Using?
WordPress Version: 5.6-RC1-49667PHP Version: 7.3.23MySQL Version: 5.6.19Caldera Forms Version: 1.9.2WP_DEBUG: 1

## Does Your Issue Persist When You Disable All Other Plugins and Switch To The Default Theme?
Yes.

## What Is The Unexpected Behaviour?
Clicked on "Export" and the screen just darkens without anything happen after that.

## What PHP Errors Have You Logged While Reproducing This Bug?

No PHP errors, but a JS error.

## What JavaScript Errors Have You Seen While Reproducing This Bug?

Uncaught TypeError: d.triggerClass is undefined
    params http://example.de/wp-content/plugins/caldera-forms/assets/build/js/wp-baldrick-full.min.js?ver=1.9.2:3
    c http://example.de/wp-content/plugins/caldera-forms/assets/build/js/wp-baldrick-full.min.js?ver=1.9.2:2
    baldrick http://example.de/wp-content/plugins/caldera-forms/assets/build/js/wp-baldrick-full.min.js?ver=1.9.2:3
    jQuery 9
    baldrick http://example.de/wp-content/plugins/caldera-forms/assets/build/js/wp-baldrick-full.min.js?ver=1.9.2:3
    jQuery 2
    baldrick http://example.de/wp-content/plugins/caldera-forms/assets/build/js/wp-baldrick-full.min.js?ver=1.9.2:3
    baldrickTriggers http://example.de/wp-content/plugins/caldera-forms/assets/build/js/admin.min.js?ver=1.9.2:1
    <anonymous> http://example.de/wp-content/plugins/caldera-forms/assets/build/js/admin.min.js?ver=1.9.2:1
    jQuery 13
        e
        t
        setTimeout handler*l/<
        c
        fireWith
        fire
        c
        fireWith
        ready
        B
        EventListener.handleEvent*
        <anonymous>
        <anonymous>

Zodiac1978:
Not sure if this could be related to jQuery update:
https://make.wordpress.org/core/2020/11/05/updating-core-jquery-to-version-3-part-2/

New0:
To add precision to the comment at https://github.com/CalderaWP/Caldera-Forms/issues/3642#issuecomment-746309213 in this issue.

[...]

PR #3651, opened 2020-12-17
ReEnqueue jquery on main admin page

New0:
For #3642 #3639 

@Shelob9 I re-enqueued jQuery in a straightforward way from the assets folder in order to remove it when we remove assets and we don't need to load an unminified version.


stuart-sequeira (approved):
Reviewed and approve the changes.  I know that some of the tests aren't building and are 'failing' but it is a false negative so far as it is not the actual tests failing, but rather the build process.

Issue #3652, opened 2020-12-18
Color Picker issue on WP >= 5.6


New0:
Test at Ghost Inspector - https://app.ghostinspector.com/ - showed an issue for the color picker field when WP installed is >= 5.6

It looks like a portion of code targetting old IE browsers < 9 is causing this issue.

New0:
Closed via - https://github.com/CalderaWP/Caldera-Forms/commit/cf1e82a02e65adaedb74106fcd54ea6721294d7e -

PR #3653, opened 2020-12-18
Remove color Picker field support for IE < 9 browsers

New0:
For #3652 

Updating the Color Picker code creates new issues, removing the portion of code targetting old ie browser < 9 fixes the color picker field on current browsers.


stuart-sequeira (approved):
Reviewed and approved.  Noted removed of support for IE<9

PR #3656, opened 2020-12-21
Add 'permission_callback' => TRUE

deputt:
Adding 'permission_callback' => TRUE to line 62 (also in crud.php line 25) seems to stop REGISTER_REST_ROUTE notice.


New0:
Thank you @deputt

PR #3657, opened 2020-12-21
Add 'permission_callback' => TRUE

deputt:
Adding 'permission_callback' => TRUE to line 25 (also in tokens.php line 62) seems to stop REGISTER_REST_ROUTE notice.


Shelob9:
@deputt Thanks for the PR. I editted this to be against the develop branch, which worked fine. We only accept PRs against develop -- we then merge develop to master, once per release.

Shelob9:
@deputt Thanks for this PR and  for #3657

@New0 Can we put these two into the next release please?

classes/api/tokens.php
```diff
@@ -22,6 +22,7 @@ public function add_routes( $namespace ) {
 			array(
 				'methods'         => 'post',
 				'callback'        => array( $this, 'get_new_nonce' ),
+			        'permission_callback' => TRUE,
```
Shelob9, line 25: This should be lower case, but that's not a huge deal.

Shelob9 (approved):
I'm not sure what's up with tests, but otherwise, approved.

Issue #3662, opened 2020-12-31
Stripe processor makes the "Save Form" Button disappear


New0:
On the Form edition page, after adding a Stripe processor to the Form and reloading the page, the "Save Form " button doesn't appear.
![Screenshot_2020-12-31 Caldera Forms Admin ‹ VSLAN — WordPress](https://user-images.githubusercontent.com/16348693/103422923-2a388900-4ba4-11eb-8081-97c881fc0c85.png)

omniacode:
Same thing is happening with the Authorize.net Addon.

New0:
This issue should be fixed in 1.9.4 - https://downloads.wordpress.org/plugin/caldera-forms.1.9.4.zip - don't hesitate to reopen this issue if this wasn't the case.

Thank you for reporting this issue and for your patience.

PR #3663, opened 2020-12-31
Extend jQuery 12 script to all CF Pages in admin

New0:
For #3662 

This help restore the "Save Form" button on the Form edition page when a Stripe processor is in use.


kevin39:
Still broken with calderawp_license_manager

New0:
Replaced by #3674

Issue #3671, opened 2021-01-15
Save Button and conditions edits unavailable in some forms depending on processors.


KilakwaBT:
We have upgraded to ver 1.9.3 with WP 5.6. While this new version has solved some of the jQuery issues there is still an issue with some forms. When specific processors are added to the form. The Save button and the ability to add and edit conditions are not available when opening for editing. We have identified the following processors as causing this.

- User Registration
- User login
- Stripe

There may be others we have yet to identify. 

WordPress Version: 5.6PHP Version: 7.3.26MySQL Version: 5.7.32Caldera Forms Version: 1.9.3WP_DEBUG: 1

jackssn:
I have same problem on my very old form with email confirmation. And the problem appeared after update Wordpress (and plugins)

In debugger firefox I see empty button:

`<div id="caldera-header-save-button"></div>` by event in `/wp-includes/js/dist/vendor/react-dom.min.js?ver=16.13.1`

OldItalian:
I've had to result to loading older versions of jQuery using Enable jQuery Migrate Helper. Many people seem to be in the same boat. Hopefully 1.9.4 is released to fix issues with WP 5.6 and jQuery compatibility.

racemaniacs:
I had the same issue using the auto-responder processor.
After updating to 1.9.4-b1 the problem seems to be resolved

kevin39:
Same issue here. All is up-to-date but "Save form" missing. I had to  load older jQuery version.

kevin39:
Oops.
Just saw 1.9.4-b1 is available : https://github.com/CalderaWP/Caldera-Forms/releases/tag/1.9.4-b1

All is fine now ! I restored the official jQuery version and disable "jQuery Migrate Helper".

Upperholme:
1.9.4-b1 didn't work for me. I installed it and then couldn't navigate through the tabbed settings for my forms - lots of red alerts in my browser console. Sticking with 1.9.3 and implementing the legacy jQuery with the Enable jQuery Migrate Helper plugin got me a result. Interesting only one of the forms on my site - the first to be created - seems to be affected.

nixonwhite:
I can say that 1.9.4-b1 didn't work for me. jQuery Migrate Helper is also useless.

ericger:
the beta let me work on the forms but makes my fusion builder useless. Can't open anything anymore. and when you don't deactivate your Paypal pro plugin before installing the beta, you are screwed.

[...]

Issue #3668, opened 2021-01-10
Can't Create Forms


heykymmie:
Hi, I'm using Version 1.8.3 of Caldera Forms. Wordpress is Version 5.6.1. I can't clone nor create a new form. Help!

New0:
This issue should be fixed in 1.9.4 - https://downloads.wordpress.org/plugin/caldera-forms.1.9.4.zip - don't hesitate to reopen this issue if it is not the case.

Thank you for reporting this issue and for your patience.

PR #3674, opened 2021-02-22
Fix wp 5.6 process conflicts

New0:
For #3671 -  #3668 - #3669

- Form edition page save button is failing on WP 5.6+ when processors like stripe are in the form.
- Adding a form seems not to be triggered at all on WP 5.6
- Clicking View entry button on an entry in the entries list doesn't show the entry.

To make it work those are the changes introduced \:
- added a condition to make sure the field edition process ( in @calderajs package ) is passed all data required
- dequeue jQuery 1.12.4
- changed deprecated jQuery events
- call jQuery instead of $
- add a click event to reTrigger baldrickTrigger events ( jQuery update seem not to listen to baldrick events for DOM elements that are not shown on the first pageload )
- add an onClick event to call baldrickTrigger on the view entry button


riverBanjo:
@New0 I was able to confirm this branch was able to save with a Stripe Processor on a form. I can also view an entry. I can not edit an entry though.

stuart-sequeira (approved):

Issue #3691, opened 2021-11-16
Submissions get deleted when a new condition is added


riverBanjo:
## What Version Of Caldera Forms, WordPress and PHP Are You Using?
WordPress Version: 5.8.2
PHP Version: 7.4.1
MySQL Version: 8.0.16
Caldera Forms Version: 1.9.5

## Does Your Issue Persist When You Disable All Other Plugins and Switch To The Default Theme?
Yes

## What Is The Unexpected Behaviour?
If you have entries on a form and then add a new condition the entries will be deleted.

## What PHP Errors Have You Logged While Reproducing This Bug?
No PHP error logs.

## What JavaScript Errors Have You Seen While Reproducing This Bug?
No JS errors I could see.

riverBanjo:
Also occurs on WordPress Version: 5.8.2 PHP Version: 7.4.25 MySQL Version: 5.7.32 Caldera Forms Version: 1.9.5

Upon further investigation the entries are deleted from the database along with entry values.

New0:
Fixed via #3692

PR #3692, opened 2021-12-03
Remove precise version of  "symfony/polyfill-mbstring" dependency

New0:
For #3691 

It seems like the limitation of the  "symfony/polyfill-mbstring" package to version "1.20" is creating this issue. This version was set to prevent an error during the SVN process as the newer version held not compatible files, further testing is required for this PR to be complete.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -7,7 +7,7 @@
         "forms",
         "caldera"
     ],
-    "license": "GPL-2.0+",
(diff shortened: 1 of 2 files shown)


New0:
@riverBanjo I could update WordPress.org trunk/ folder as expected. The failing tests come from the docker environment maintenance. GI tests all passed.

Let me know if that fines the issue manually testing.

